PR #12, opened 2014-12-18
Add fetch to task arguments.

keathley:
This change allows TaskArguments to respond to fetch in the same way that a hash would.  I ran into this scenario where I was trying to pass the args for a rake task directly into a class constructor.  Calling args.fetch in that case results in nil values being returned.  This was confusing at first because when the TaskArguments object is 'puts' or inspected, it returns what looks like a hash.  It wasn't until I checked the class of the args that I realized the discrepancy.  This change helps eliminate that confusion and would allow the arguments to be passed in directly.

I'm not sure if this is the right approach or if you should just call `to_hash` on the args (which is what I ended up doing).  Any feedback would be great.  Thanks!

### DIFF
--- a/lib/rake/task_arguments.rb
+++ b/lib/rake/task_arguments.rb
@@ -83,8 +83,8 @@ module Rake
       @hash.has_key?(key)
     end
 
-    def fetch(key, &block)
-      @hash.fetch(key, &block)
+    def fetch(*args, &block)
+      @hash.fetch(*args, &block)
     end
 
     protected
--- a/lib/rake/task_arguments.rb
+++ b/lib/rake/task_arguments.rb
@@ -83,6 +83,10 @@ module Rake
       @hash.has_key?(key)
     end
 
+    def fetch(key, &block)
+      @hash.fetch(key, &block)
+    end
+
     protected
 
     def lookup(name) # :nodoc:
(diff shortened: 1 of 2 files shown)


lib/rake/task_arguments.rb
```diff
@@ -83,6 +83,10 @@ def has_key?(key)
       @hash.has_key?(key)
     end
 
+    def fetch(key, &block)
```
cantino: `fetch` can also take a second argument as a default.  Perhaps something like:

```ruby
def fetch(*args, &block)
  @hash.fetch(*args, &block)
end
```

keathley: I've pushed up a fix to this.  Thanks for catching that!

test/test_rake_task_arguments.rb
```diff
@@ -124,4 +131,10 @@ def test_extra_args_with_less_than_named_arguments
     assert_equal [], ta.extras
   end
 
+  private
+
+  def older_ruby?
```
hsbt: I dropped old ruby support at Rake 11. Could you remove this method and needless condition at L32?

hsbt: I removed this at https://github.com/ruby/rake/commit/95f212c5171dbc366e32ec56d9af61492728125d

PR #18, opened 2015-01-09
Multitask output is interleaved

randycoulman:
When running multitasks, any output produced by the tasks can be interleaved incorrectly.  If you're using Rake to do something like compile programs in other languages, the result can be error messages that no longer match the regex used by your IDE or editor, making it hard to navigate to the errors.  I'm sure there are other situations where this interleaved output causes problems as well.

For example, the following Rakefile (written to make the problem really obvious):

```ruby
25.times do |n|
  child_task_name = "child\#{n}"
  task child_task_name do
    print "Run ", "child", n, "\n"
  end
  multitask default: child_task_name
end
```

results in output that looks something like this:

```
Run child0
Run child1
Run child2
Run child3
Run childRun Run child6Run child8
Run child11Run child15
Run child19Run child23Run child7
Run childRun child13Run child16Run child20
4


Run child10Run childRun child17
Run childRun child
Run child9


Run child18

child5
14
Run child24
22
12
21
```

I've turned this Rakefile into a functional test, but the test does not fail.  It seems like the rake process launched in the test runner does not run multiple threads/processes like it should, or perhaps `popen3` is serializing the output for us.  I'd appreciate some thoughts on how to make this test fail like it should.

In order to fix the underlying problem, I'd like to introduce a little wrapper class has a `Mutex` and use it to synchronize all IO operations.

One option would be to wrap `$stdout`, `$stderr`, and any stream passed to `trace_on` at startup and when `trace_on` is called.  A further option would be to only do this if the options are such that multitasks are possible.

Another option would be to do this wrapping temporarily in `invoke_prerequisites_dynamically` and then undo it when that's done.

I'm a bit worried that always synchronizing output in this way will slow down Rake jobs, but I'd also like to fix the interleaved output.

I'd appreciate some thoughts and direction on which way to proceed here.

### DIFF
--- a/lib/rake/application.rb
+++ b/lib/rake/application.rb
@@ -7,6 +7,7 @@ require 'rake/thread_pool'
 require 'rake/thread_history_display'
 require 'rake/trace_output'
 require 'rake/win32'
+require 'rake/synchronized_output'
 
 module Rake
 
@@ -73,9 +74,11 @@ module Rake
     # call +top_level+ to run your top level tasks.
     def run
       standard_exception_handling do
-        init
-        load_rakefile
-        top_level
+        # synchronize_output_streams do
+          init
+          load_rakefile
+          top_level
+        # end
       end
     end
 
@@ -115,6 +118,7 @@ module Rake
       yield
 
       thread_pool.join
+
       if options.job_stats
         stats = thread_pool.statistics
         puts "Maximum active threads: #{stats[:max_active_threads]} + main"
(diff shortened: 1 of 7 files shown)


randycoulman:
Oh, interesting... The new test fails on jruby, but not on MRI.

randycoulman:
One other implementation note: the wrapper class could be written better, using `BasicObject`, `public_send`, and `respond_to_missing?`, but all of those methods were introduced in 1.9 and Rake is still maintaining compatibility with 1.8.7.

tenderlove:
What Ruby are you using to get the jumbled output?  I can't seem to get it on 2.3.0.

randycoulman:
@tenderlove On my MBP, every ruby I've tried shows the problem (2.2, 2.1.5, 1.9.3, JRuby).  On my iMac at home, I can't get any of them to show the problem.  I don't know how to explain the difference.  The iMac's processor is faster (3.1 GHz vs 2.5 GHz) and the MBP has an SSD.

tenderlove:
Oh, I forgot to mention on your blog post, you should try using IO#reopen, and a pipe in order to intercept all IO.  Something like this:

```ruby
rd, wr = IO.pipe
original_stdout = IO.new($stdout.to_i)
$stdout.reopen(wr)
```

Then in a thread you can read from `rd` and write to `original_stdout`.  I'm still not sure how you could coordinate between threads though.

randycoulman:
Realistically, I'm not going to have time to get back to this and figure it out any time soon, so I'll close it for now.

If anyone wants to pick this up in the future, see @tenderlove's comments on my [related blog post](http://randycoulman.com/blog/2015/01/27/need-help-with-pull-request/) (and above as well).  Also, see [this tweet from @drbrain](https://twitter.com/drbrain/status/560490073657860097).

PR #19, opened 2015-01-10
Offer alternatives if rake task not found

djberg96:
If your rake task name happens to be wrong on the command line, the output is not that helpful. Next thing you know you're grepping through rake -T output.

This pull request does some simple matching to offer alternatives if the name you specified isn't found. For example, if you have tasks "foobar" and "foobaz", but accidentally type "rake foo" on the command line, you'll see this:

    Don't know how to build task 'foo'. Did you mean one of these?

      foobar
      foobaz

### DIFF
--- a/lib/rake/task_manager.rb
+++ b/lib/rake/task_manager.rb
@@ -59,7 +59,20 @@ module Rake
       self.lookup(task_name, scopes) or
         enhance_with_matching_rule(task_name) or
         synthesize_file_task(task_name) or
-        fail "Don't know how to build task '#{task_name}'"
+        build_task_fail(task_name)
+    end
+
+    # If a task match wasn't found, try to offer suggestions.
+    # The task name must be at least 2 characters.
+    def build_task_fail(task_name)
+      msg = "Don't know how to build task '#{task_name}'"
+      if task_name.size > 1
+        matches = @tasks.keys.find_all{ |e| e =~ /#{task_name}/i }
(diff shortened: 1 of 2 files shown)


jrafanie:
:heart:  This reminds me of https://github.com/rails/rails/pull/15497.  It would be nice if it handled typos of fobar/foobr but this is much simpler and much better than what we have right now.  Note, I often typo destroy as destory so I'm partial to that rails PR.  Nevertheless, this is awesome!

lib/rake/task_manager.rb
```diff
@@ -59,7 +59,20 @@ def [](task_name, scopes=nil)
...
-        fail "Don't know how to build task '#{task_name}'"
+        build_task_fail(task_name)
+    end
+
+    # If a task match wasn't found, try to offer suggestions.
+    # The task name must be at least 2 characters.
+    def build_task_fail(task_name)
+      msg = "Don't know how to build task '#{task_name}'"
+      if task_name.size > 1
+        matches = @tasks.keys.find_all{ |e| e =~ /#{task_name}/i }
```
cantino: You probably want `/#{Regexp.escape task_name}/`

djberg96: I've never seen one in the wild, but it's possible I suppose.

hsbt:
I hope to implement this feature with did_you_mean gem.

PR #24, opened 2015-02-17
Convert manual page to use semantic -mdoc macros.

bentley:


hsbt:
@bentley Can you added "rake 10.3.2" and "Rake User Commands" sections?

hsbt:
and https://github.com/ruby/rake/commit/c34d9e0afbaee050a4202d045a42efb8aae65117

PR #29, opened 2015-02-21
Provide "Did you mean? task_name" suggestions if task was not found

yuki24:
similar to #19, but also suggests task names based on the edit distance.

### DIFF
--- a/lib/rake/task_manager.rb
+++ b/lib/rake/task_manager.rb
@@ -59,8 +59,28 @@ module Rake
       self.lookup(task_name, scopes) or
         enhance_with_matching_rule(task_name) or
         synthesize_file_task(task_name) or
-        fail "Don't know how to build task '#{task_name}'"
+        build_task_fail(task_name)
     end
+
+    # If a task match wasn't found, try to offer suggestions.
+    # The task name must be at least 2 characters.
+    def build_task_fail(task_name)
+      msg       = "Don't know how to build task '#{task_name}'"
+      threshold = (task_name.size * 0.3).ceil
+
+      matches = @tasks.keys.
+        map {|name| [levenshtein_distance(name.downcase, task_name), name] }.
+        select {|distance, _| distance <= threshold }.
+        sort.
+        map(&:last)
+
+      if matches.size > 0
+        msg += ".\nDid you mean one of these? #{matches.join(", ")}"
+      end
+
+      fail msg
+    end
+    private :build_task_fail
 
     def synthesize_file_task(task_name) # :nodoc:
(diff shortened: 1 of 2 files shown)


lib/rake/task_manager.rb
```diff
@@ -301,6 +321,33 @@ def get_description(task)
       desc
     end
 
+    # Return a value representing the "cost" of transforming str1 into str2
+    def levenshtein_distance(str1, str2)
```
zzak: We pass this method around, but I think it should be public in RubyGems, no?

yuki24: It is. We can just `include Gem::Text`and call `#levenshtein_distance`. I think you already know it had a bug, but maybe we shouldn't worry about it?

zzak: On the other hand, Rake seems to keep a safe distance from RubyGems (maybe for a reason).

This _could_ introduce some circular dependencies by depending on Gem itself.. and I'm not seeing anywhere besides `testtask` that uses it.

One could argue this method should be in a separate gem that both can require, and other people would find it useful. If that is the case, why not move it to core?

yuki24: Yeah, I've been thinking about it - providing `levenshtein_distance` as a stdlib of Ruby. Rubygems and Bundler already (and possibly Rake) use the Levenshtein algorithm so I think it's a good idea to move it to core. I'll file a feature request to https://bugs.ruby-lang.org.

That being said, I realized that rake still needs to have support for 1.8.7, meaning we can't just use fresh features introduced recently. Should we just not merge this for now? I don't feel comfortable forcing others to maintain code that is copy-and-pasted...

zzak: We should avoid duplication whenever possible

yuki24: filed a ticket: https://bugs.ruby-lang.org/issues/10882

drbrain: I don't see why we can't merge this now while we wait for ruby to have levenshtein distance.

zzak: If we merge this now I think we need to consider using `Gem::Text` instead of copy&paste

drbrain: The necessary methods for `Gem::Text` don't exist for all versions of RubyGems that rake supports.

yuki24:
closing this pull request as there's an ongoing discussion on https://bugs.ruby-lang.org/issues/11252. We might be able to use some of the features of `did_you_mean` in the near future, but I guess it's still too early to have this conversation here.

PR #30, opened 2015-02-23
Remove duplicates from FileList

BertramScharpf:
I work with TeX and the pattern %w(*.log texput.*)
catches "texput.log" twice, which results in an error
when removing the file inside a :clean task.

### DIFF
--- a/lib/rake/file_list.rb
+++ b/lib/rake/file_list.rb
@@ -198,7 +198,8 @@ module Rake
 
     def <<(obj)
       resolve
-      @items << Rake.from_pathname(obj)
+      i = Rake.from_pathname(obj)
+      @items.push i unless @items.include? i
       self
     end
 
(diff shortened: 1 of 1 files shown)


hsbt:
At least you fixed test fail.

BertramScharpf:
It's not me who wrote the bug.

yuki24:
@BertramScharpf That's you who [broke the CI build](https://travis-ci.org/ruby/rake/builds/51836027). 3 tests are failing. Fix the bug you are trying to fix without breaking existing tests.

BertramScharpf:
Then just regard it as a bug report. There is still a "--no-commit" option in the Git merge command.

As long as such a silly bug can occur I will not be convinced of unit tests. Even, I do not want to participate in the kind of programming style I have seen.

PR #34, opened 2015-03-13
Make FileList#exclude more analogous to FileList#include

epidemian:
This PR changes FileList so that `a_filelist.exclude(something)` is more consistent with `a_filelist.exclude(*FileList[something])`.

Now `FileList#exclude` can accept curly brace patterns, like `"{foo,bar}.rb"`, the same way `FileList#include` does.

And it now also accepts arrays, or array-like things, excluding each element of that array.

### DIFF
--- a/lib/rake/file_list.rb
+++ b/lib/rake/file_list.rb
@@ -151,7 +151,11 @@ module Rake
     #
     def exclude(*patterns, &block)
       patterns.each do |pat|
-        @exclude_patterns << Rake.from_pathname(pat)
+        if pat.respond_to? :to_ary
+          exclude(*pat.to_ary)
+        else
+          @exclude_patterns << Rake.from_pathname(pat)
+        end
       end
       @exclude_procs << block if block_given?
       resolve_exclude unless @pending
--- a/lib/rake/file_list.rb
+++ b/lib/rake/file_list.rb
@@ -369,7 +369,10 @@ module Rake
         when Regexp
           fn =~ pat
         when GLOB_PATTERN
-          File.fnmatch?(pat, fn, File::FNM_PATHNAME | File::FNM_EXTGLOB)
+          flags = File::FNM_PATHNAME
+          # Ruby <= 1.9.3 does not support File::FNM_EXTGLOB
+          flags |= File::FNM_EXTGLOB if defined? File::FNM_EXTGLOB
+          File.fnmatch?(pat, fn, flags)
         else
           fn == pat
         end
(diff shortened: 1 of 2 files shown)


epidemian:
A bit of background: i came upon these oddities in `FileList#exclude` when wanting to split a bit test suite that we have into two, lets say: the important tests, and the unimportant tests (which we don't care too much if they break). All the tests in under test/foo and test/bar and some specific tests under test/baz are important; the rest are unimportant. So i wanted to do something like this:

```ruby
Rake::TestTask.new do |t|
  t.test_files = FileList['test/**/test_*.rb']
end

important_tests = %w[
  test/{foo,bar}/**/test_*.rb
  test/baz/test_something_important_about_baz.rb
  test/baz/test_something_else_of_importance.rb
]
Rake::TestTask.new('test:important') do |t|
  t.test_files = FileList[important_tests]
end
Rake::TestTask.new('test:unimportant') do |t|
  t.test_files = FileList['test/**/test_*.rb'].exclude(important_tests)
end
```

This way, if we decide some unimportant test should be promoted to `test:important` or an important test be demoted to `test:unimportant`, we can just change that one `important_tests` array and have both tasks updated.

With the changes in this PR, that code should run correctly :smiley_cat:

epidemian:
Ups. The Travis build failed on Ruby 1.8.7 and 1.9.3. It seems `File::FNM_EXTGLOB` was added on 2.0.0 :tired_face:

I can see if i can make it work with those versions of Ruby (any pointers are welcomed!), but i'd like to know if these proposed changes make sense. Does the use case described above sound reasonable?

epidemian:
Well, i went ahead and simply added a check for the existence of `File::FNM_EXTGLOB` before using it; and skipped the test for brace pattern exclusion on rubies that don't implement that constant.

Another approach could be doing the brace pattern expansion manually if `File::FNM_EXTGLOB` is not defined (we could copy [Rubinius' implementation](https://github.com/rubinius/rubinius/commit/d95487ca490e775eabf7eef499202182bd640bc8), which is quite nice, and plain Ruby), but i didn't want to bloat FileList with backward Ruby support stuff. If you think that would be the better option though, i'll be glad to implement it that way, and have a more consistent behaviour of `FileList#exclude` across Ruby versions :smiley_cat:

hsbt:
@epidemian Sorry, my late response. Your usecase ( https://github.com/ruby/rake/pull/34#issuecomment-78821782 ) is reasonable to me. I will release at Rake 11 with this pull request.

epidemian:
Great! Thanks @hsbt :smile_cat:

PR #38, opened 2015-04-11
Use append (+) instead of bitwise or for significant performance enhancement

42races:
Use append (+) instead of bitwise or for significant performance enhancement
Simplify, the arg_description method

Signed-off-by: Deepak Kumar <deepakkumarnd@gmail.com>


42races:
closing because of build failure

PR #39, opened 2015-04-11
Chained extensions

pjump:
rules such as 

``` ruby
rule ".min.js" => ".js" do |t|
end
```

didn't work because Rake would simple replace the js part of".min.js" with ".js", creating a cyclic dependency. This PR makes rules like this work by always making sure the whole target pattern of is replaced, rather than just the very final extension. 

All tests green. A test for the new feature added.

### DIFF
--- a/test/test_rake_rules.rb
+++ b/test/test_rake_rules.rb
@@ -10,6 +10,7 @@ class TestRakeRules < Rake::TestCase
   OBJFILE    = "abc.o"
   FOOFILE    = "foo"
   DOTFOOFILE = ".foo"
+  MINFILE = 'abc.min.o'
 
   def setup
     super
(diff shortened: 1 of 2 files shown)


pjump:
On a somewhat unrelated topic, does anyone know how to gemify this repo so I can install it?

pjump:
Any opinion on this?

hsbt:
@pjump Sorry to my late response. I agreed to merge your proposal.

test/test_rake_rules.rb
```diff
@@ -10,6 +10,7 @@ class TestRakeRules < Rake::TestCase
   OBJFILE    = "abc.o"
   FOOFILE    = "foo"
   DOTFOOFILE = ".foo"
+  MINFILE = 'abc.min.o'
```
hsbt: Can you align `=` indentation?

PR #48, opened 2015-05-30
Remove unnecessary monkey patching

pabloh:
This kind of monkey patching isn't really necessary since `Time#<=>` is smart enough to try inverting the operands if the argument it receives is not an instance of Time.

Also the alias chain can be a source of headaches if you want to use it simultaneously with the new method `Module#prepend`: https://github.com/rails/rails/pull/19878/#issuecomment-106851263


hsbt:
It seems good because travis-ci is green.

PR #49, opened 2015-06-03
custom test loader: t.load = "file://path/to/loader"

tmornini:
This code allows users of Rake::TestTask to use custom test loaders.

This was a required for an experiment we've been doing at Subledger that required the test files to be loaded rather than required due to non-'.rb' filename extensions.

When implementing this, I noticed that there was no else-clause in the case statement in #load so I added one.

Wrote some tests for the custom loader option (pass a file:// URI string to loader) and the else-clause too.

Anxious to hear your feedback.

### DIFF
--- a/lib/rake/testtask.rb
+++ b/lib/rake/testtask.rb
@@ -57,9 +57,10 @@ module Rake
 
     # Style of test loader to use.  Options are:
     #
-    # * :rake -- Rake provided test loading script (default).
+    # * :rake   -- Rake provided test loading script (default).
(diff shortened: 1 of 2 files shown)


tmornini:
Only ruby-head is failing on Travis, and the failures don't appear to be anywhere near my changes.

tmornini:
Rebased this on master and ruby-head is now broken, but nowhere near the changes in this PR. :-)

tmornini:
PIng

tmornini:
@hsbt Sure thing, I'll get it done this weekend

PR #55, opened 2015-07-02
Add a :triggers option when defining rake tasks to run tasks after the task is run

jsm:
Sometimes I want a task to trigger other tasks AFTER it's been run, rather than before, like a prerequisite.

This functionality is completely backwards compatible.

For example:

```
task default: [:p1, :p2], triggers: [:t1, :t2] do
  puts 'default'
end

[:p1, :p2, :t1, :t2].each do |t|
  task t do
    puts t
  end
end
```
would output
```
p1
p2
default
t1
t2
```

### DIFF
--- a/lib/rake/task.rb
+++ b/lib/rake/task.rb
@@ -15,6 +15,9 @@ module Rake
     # List of prerequisites for a task.
     attr_reader :prerequisites
 
+    # List of task triggers for a task.
+    attr_reader :triggers
+
     # List of actions attached to a task.
     attr_reader :actions
 
@@ -50,13 +53,13 @@ module Rake
 
     # List of prerequisite tasks
     def prerequisite_tasks
-      prerequisites.map { |pre| lookup_prerequisite(pre) }
+      prerequisites.map { |pre| lookup_task(pre) }
     end
 
-    def lookup_prerequisite(prerequisite_name) # :nodoc:
-      application[prerequisite_name, @scope]
+    def lookup_task(task_name) # :nodoc:
+      application[task_name, @scope]
     end
-    private :lookup_prerequisite
+    private :lookup_task
 
     # List of all unique prerequisite tasks including prerequisite tasks'
     # prerequisites.
@@ -76,6 +79,11 @@ module Rake
     end
     protected :collect_prerequisites
 
+    # List of triggered tasks
+    def triggered_tasks
+      triggers.map { |trigger| lookup_task(trigger) }
+    end
+
     # First source from a rule (nil if no sources)
     def source
       sources.first
@@ -86,6 +94,7 @@ module Rake
     def initialize(task_name, app)
       @name            = task_name.to_s
       @prerequisites   = []
+      @triggers        = []
       @actions         = []
       @already_invoked = false
       @comments        = []
@@ -96,9 +105,10 @@ module Rake
       @locations       = []
     end
 
-    # Enhance a task with prerequisites or actions.  Returns self.
-    def enhance(deps=nil, &block)
+    # Enhance a task with prerequisites, triggers or actions.  Returns self.
+    def enhance(deps=nil, triggers=nil, &block)
       @prerequisites |= deps if deps
+      @triggers |= triggers if triggers
       @actions << block if block_given?
       self
     end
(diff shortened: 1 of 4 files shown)


jsm:
Can I get some feedback please?

drbrain:
Why not add blocks to the default task?

```ruby
$ cat t.rake
task :p1 do puts :p1 end
task :p2 do puts :p2 end

task default: [:p1, :p2] do
  puts 'default'
end

task :default do
  puts :t1
end

task :default do
  puts :t2
end

$ rake -f t.rake
p1
p2
default
t1
t2
```

jsm:
Because they might be tasks I want to be able to run separately or triggered by different task.

like 

task :hello, triggers: [:world]
task :goodbye, triggers: [:world]

drbrain:
If you want to run them separately why not make them regular tasks with regular dependencies and document which task is the correct task to run to accomplish whatever it is they wish to do?

A second option is to place the logic in a method in a library and use rake to declare the necessary dependencies.  This way you can add extra work to `default` above and also add it to a separate task that is run independently.

Rake's primary purpose is to help a user structure a set of dependent tasks so they can think about and document them clearly.

I feel that adding a secondary psuedo-dependency system to rake detracts from this purpose.

I feel you can use existing rake features to construct equivalent functionality without triggers.

jsm:
A few things that I'm looking for that your proposals don't cover:

1. Multitasking. What if i want to run A, B, C in parallel, then trigger my task, then after it's done, I want to make sure X, Y, X run in parallel?
2. Tracking invocation chains. The reason I added this was because I wanted to be able to get the invocation chain of a task that is called during/after task execution. The only way to execute another rake task right now is to do Rake::Task['task'].invoke, but this does with with a clean invocation chain.

jsm:
One possibility would be something like

```
task my_task: ['my_task:pre', 'my_task:inner', 'my_task:post']
namespace :mytask
  multitask pre: [:p1, :p2]
  
  task :inner do
    # Normal task logic here
  end

  multitask post: [:t1, :t2]
end
```

But that's a lot of complexity as opposed to

```
multitask my_task: [:p1, :p2], triggers: [:t1, :t2] do
  # Task logic here
end
```

drbrain:
Currently rake supports only one direction of dependency, from a task to the tasks that need to run before it.  Your patch may allow you to omit one task name when building the dependency graph, but you still build the same graph with rake as-is or with your patch.  Since the topology of the tasks is the same with and without your patch, I prefer to leave rake as-is.

Here is an example using rake as-is to run tasks in parallel before and after a middle coordinating task:

```ruby
def work m
  sleep rand
  puts "#{m} #{Thread.current.object_id}"
end

task :a do work :a end
task :b do work :b end
task :c do work :c end

task x: :middle do work :x end
task y: :middle do work :y end
task z: :middle do work :z end

multitask middle: %w[a b c] do
  work :middle
end

multitask finish: %w[x y z] do
  work :finish
end
```

```
$ rake -f t.rake finish
b 70192742881740
c 70192742880740
a 70192734503480
middle 70192734798340
z 70192734504780
y 70192734506000
x 70192734798340
finish 70192734798340
```

Here you can see the single-letter tasks operate in parallel on separate threads.  Running this will give different orders as the single-letter tasks complete in different amounts of time.

Compared to your terse example:

```ruby
multitask my_task: [:p1, :p2], triggers: [:t1, :t2] do
  # Task logic here
end
```

My example contains one more name (`finish`) but the topology is the same.  Through your patch `triggers` creates an implied dependency to tasks `t1` and `t2`.

my example topology

my_task

Your patch doesn't make a set of tasks easier to think about.  Instead of considering one direction of edge (from a task to its dependencies) you must consider two directions for an edge.  This makes the Rakefile harder to reason about while remaining topologically identical.

Additionally, your patch doesn't support the introspection capabilities (`rake -P`) that allow easy debugging of rake (used by [graph](https://rubygems.org/gems/graph) to outputs a diagram of your dependencies).

Being more terse in the declaration of tasks at the expense of implementation complexity (more maintenance for me, especially if you were to support `rake -P`) and added mental burden (for users of this feature) are not positive tradeoffs for myself or rake users.

jsm:
Makes sense. Thanks a lot for your in-depth feedback.

jsm:
I just realized how quickly the timestamps say I read that and closed this after your response. I want to assure you that I did read it very closely and thought about it.

drbrain:
No worries!

I do like that your patch allows you to create that implied task without naming it, but I feel the cost is too high.

PR #56, opened 2015-07-12
Improve markup by modifying a linux command of 'CD' into lowercase 'cd' (enclosed by '+')

shunsuke227ono:
I changed CD to `cd` (enclosed by `+`) in CONTRIBUTING.rdoc.

I guess CD in CONTRIBUTING.rdoc is supposed to mean a linux `cd` command, but if it is written in uppercase, it might be confusing because it is normally written in lowercase(http://www.linfo.org/cd.html).

It is a subtle change but to make it a little bit easier to understand.


hsbt:
Thanks! I confirmed test fail, but it's not related your commit.

PR #64, opened 2015-08-28
Convert manual page to use semantic -mdoc macros.

bentley:
@hsbt re #24 and #63. This re‐adds the “rake 10.4.2” section. I don’t see a way to re‐add “Rake User Commands”; is it really necessary?


hsbt:
> is it really necessary?

I'm ok to remove it.

PR #67, opened 2015-09-09
Fix TestTask verbose

mblumtritt:
`rake test —verbose` did not result in verbose mode for MiniTest.

The verbose option needs to be forwarded to the Ruby call.

Missing tests are added.

### DIFF
--- a/lib/rake/testtask.rb
+++ b/lib/rake/testtask.rb
@@ -128,6 +128,7 @@ module Rake
       opts = @ruby_opts.dup
       opts.unshift("-I\"#{lib_path}\"") unless @libs.empty?
       opts.unshift("-w") if @warning
+      opts.unshift('--verbose') if @verbose
       opts.join(" ")
     end
 
--- a/test/test_rake_test_task.rb
+++ b/test/test_rake_test_task.rb
@@ -10,6 +10,7 @@ class TestRakeTestTask < Rake::TestCase
     assert_equal :test, tt.name
     assert_equal ['lib'], tt.libs
     assert_equal 'test/test*.rb', tt.pattern
+    assert_equal false, tt.warning
     assert_equal false, tt.verbose
     assert Task.task_defined?(:test)
   end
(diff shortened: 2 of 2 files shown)


mblumtritt:
I run same test as in Travis and all passed successfully. Can someone explain what's wrong here, please?

MarkDBlackwell:
> rake test `--verbose` did not result in verbose mode for `MiniTest`. The verbose option needs to be forwarded to the Ruby call.

`Minitest` sometimes emits the message: "You have skipped tests. Run with `--verbose` for details."

Likely, this message stimulated the above suggestion.

[This comment](https://github.com/seattlerb/minitest/pull/564#issue-94545052) by @shunsuke227ono excellently explains what's really happening—as well as [this one](https://github.com/rails/rails/issues/13641#issuecomment-31957266) by @prathamesh-sonpatki.

Generally, this problem only arises when three independent bodies of software—`Rake::TestTask`, `Minitest`, and user test code—are combined. In that case, the proper message should suggest `TESTOPTS="-v"` (rather than `--verbose`).

Perhaps `Rake::TestTask` could detect use of `Minitest`. Then it could produce the correct message (in case tests were skipped) by subclassing [`SummaryReporter`](https://github.com/seattlerb/minitest/blob/fc96854d3bf6495ea9f7dad821bb64f6bf03440f/lib/minitest.rb#L556) to override method, [`summary`](https://github.com/seattlerb/minitest/blob/fc96854d3bf6495ea9f7dad821bb64f6bf03440f/lib/minitest.rb#L607) as `Minitest` recommends [here](https://github.com/seattlerb/minitest/blob/fc96854d3bf6495ea9f7dad821bb64f6bf03440f/lib/minitest.rb#L553). Rake might substitute these classes in this [list](https://github.com/seattlerb/minitest/blob/fc96854d3bf6495ea9f7dad821bb64f6bf03440f/lib/minitest.rb#L125).

Rake's own test suite uses `Minitest`, and it displays the problem when tests are skipped (e.g. on Windows).

Issue #22, opened 2015-01-30
Default default rake


fulldecent:
Some rakefiles don't have default tasks. A good reason would be if each of the tasks does something destructive. For them, a great default task would be to list all available tasks. Best practice is documented at http://stackoverflow.com/a/11320444/300224

For DRY across the thousands of rakefiles out that use the above default, I recommend this be the default for all scripts without a default.

Presently, running `rake` where there is no default task produces a return status of 0 and this output:

 > rake aborted!
 > Don't know how to build task 'default'
 > 
 > (See full trace by running task with --trace)

heemin32:
Good opinion. However I think listing all available tasks should be implemented using another option or another task name instead of treating it as a default.

dslh:
I think for a lot of newcomers it would be a big help just to update the error message to something like `Don't know how to build task 'default' (see --tasks)`

fulldecent:
I would be perfectly happy with @dslh's solution. I'm having trouble building at the moment, but can submit a patch.

dslh:
I know that would have saved me five minutes when I was starting out with rake :)

hsbt:
I merged https://github.com/ruby/rake/pull/76.

PR #76, opened 2015-10-12
More helpful error message for unknown tasks.

dslh:
Addresses issue #22. Makes it a bit easier for new users to get started by pointing them at the `--task` parameter if they e.g. run rake with no arguments in a project with no default task.


hsbt:
:+1:

PR #80, opened 2015-10-26
Mark the FIXNUM_MAX constant as private with an annotation

pvdb:
@hsbt _et al._ - a minor change to `lib/rake/application.rb` which tried to declare the `FIXNUM_MAX` constant as `private`, which isn't possible in Ruby... I've replaced the `private` method call with an annotation similar to the `internal` annotation on [line 144 of that same file][one_four_four], to at least express the intent.

[one_four_four]: https://github.com/ruby/rake/blob/master/lib/rake/application.rb#L144

### DIFF
--- a/lib/rake/application.rb
+++ b/lib/rake/application.rb
@@ -783,7 +783,8 @@ module Rake
       backtrace.find { |str| str =~ re } || ''
     end
 
-  private
+    # private -----------------------------------------------------------------
+
     FIXNUM_MAX = (2**(0.size * 8 - 2) - 1) # :nodoc:
 
   end


drbrain:
Constants can be made private in Ruby:

```
$ ruby -e 'class A; B = 1; end; p A::B; class A; private :B; end; p A::B'
1
-e:1:in `private': undefined method `B' for class `A' (NameError)
	from -e:1:in `<class:A>'
	from -e:1:in `<main>'
```

But not this way:

```
$ ruby -e 'class A; private; B = 1; end; p A::B'
1
```

pvdb:
Thanks for your reply, @drbrain, but I don't think the snippet actually proves what you intended, at least for the versions of Ruby I've tested it on _(`1.9.3`, `2.0.x`, `2.1.x` and `2.2.x`)_.

AFAICT, the `NameError` is NOT thrown by `p A::B` in the 4th part of the snippet, but rather by the attempted _(but unsuccessful)_ `private` declaration in the 3rd part:

```
$ ruby -e 'class A; B = 1; end; p A::B; class A; private :B; end;'
1
-e:1:in `private': undefined method `B' for class `A' (NameError)
	from -e:1:in `<class:A>'
	from -e:1:in `<main>'
```

... or even simpler:

```
$ ruby -e 'class A; B = 1; private :B; end'
-e:1:in `private': undefined method `B' for class `A' (NameError)
	from -e:1:in `<class:A>'
	from -e:1:in `<main>'
```

In the meantime, though, I learned that in version `1.9.3`, Ruby introduced `Module#private_constant`, which does the trick:

```
$ ruby -e 'class A; B = 1; end; p A.constants; p A::B; class A; private_constant :B; end; p A.constants; p A::B'
[:B]
1
[]
-e:1:in `<main>': private constant A::B referenced (NameError)
```

However, I cannot use `Module#private_constant` in this PR, as that would break compatibility of `rake` with version `1.8.7` of Ruby.

Therefor, I think my PR is still valid: the current use of `private` in `lib/rake/application.rb` doesn't actually do anything, as you confirm in your second snippet, and also it's probably more a matter of expressing the intent that `FIXNUM_MAX` should be considered a "private" constant _(which is what I tried to do with my annotation)_ then it is of strictly enforcing its `private`-ness :smiley:

drbrain:
You are right, I'm sorry.

The intent is to make the private constant, so I think the best fix would be to have `private_constant :B if respond_to? :private_constant`

pvdb:
Thanks @drbrain!

Even though your suggestion would work...

```ruby
FIXNUM_MAX = (2**(0.size * 8 - 2) - 1) # :nodoc:
private_constant :FIXNUM_MAX if respond_to? :private_constant
```

... it would make the whole thing look a bit convoluted in my opinion, and would probably require a wee bit of commentary in the code to explain what's going on and why;  what's more, it would only achieve the original intent in certain circumstances.

Considering that the constant probably doesn't belong in `lib/rake/application.rb` in the first place _(from an encapsulation point-of-view, it could be argued that the `FIXNUM_MAX` constant should not be `Rake::Application`'s to govern, as it has very little to do with that class's responsibilities)_ I'm gonna close this PR, and issue a different one that refactors the constant into a more suitable place... I'll tag you on that one for review, if that's OK?

drbrain:
Sounds good.

Yeah, it's clumsy, but portable code usually has some ugliness to it.

pvdb:
> Sounds good.

#82 ... cheers!

PR #82, opened 2015-10-29
Define `Fixnum::MAX` constant, until such time as Ruby itself does...

pvdb:
@drbrain _et al._ - this PR moves the `FIXNUM_MAX` constant out of `lib/rake/application.rb`, and puts it in a `Fixnum` extension as `Fixnum::MAX`, which @matz himself is considering adding to Ruby:

https://bugs.ruby-lang.org/issues/7517

This is - IMHO - a much more elegant solution for the constant originally introduced in jimweirich/rake#135 and discussed - at length :smile: - in #80 

The additional benefit of this implementation is that - once Ruby defines the constant itself - `rake` will pick up the "correct" value for the constant, as opposed to the one it defines itself, which may be wrong on certain platforms.

### DIFF
--- a/lib/rake.rb
+++ b/lib/rake.rb
@@ -41,6 +41,7 @@ require 'ostruct'
 require 'rake/ext/module'
 require 'rake/ext/string'
 require 'rake/ext/time'
+require 'rake/ext/fixnum'
 
 require 'rake/win32'
 
--- a/lib/rake/application.rb
+++ b/lib/rake/application.rb
@@ -452,7 +452,7 @@ module Rake
             "(default is number of CPU cores + 4)",
             lambda { |value|
               if value.nil? || value == ''
-                value = FIXNUM_MAX
+                value = Fixnum::MAX
               elsif value =~ /^\d+$/
                 value = value.to_i
               else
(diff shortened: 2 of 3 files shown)


drbrain:
This is not portable due to https://bugs.ruby-lang.org/issues/7517#note-4 so I prefer this stay in Rake::Application as a private constant.

I prefer not to add more monkey-patches to rake, the few that exist are already annoying to maintain.

pvdb:
Hey @drbrain, 

You wrote:
> This is not portable due to https://bugs.ruby-lang.org/issues/7517#note-4  

Understood, in the same way as the formula provided by @shyouhei in that bug report is wrong on certain platforms, as pointed out by @headius:

    1 << (1.size * 8 - 2) - 1

... so is the formula introduced into `rake` by @michaeljbishop in jimweirich/rake#135:

    2 ** (0.size * 8 - 2) - 1

They don't even calculate the same value :smile: (which is why I assume @matz is proposing to add `Fixnum::MAX` to Ruby itself, in a correct and portable way, to avoid having people to approximate it with demonstrably wrong, hand-cranked formulas).

However, the problem I have with `Rake::Application::FIXNUM_MAX` _(whether it's private or not)_ is that the value it calculates is wrong, and - because of its highly suggestive name - it purports to be something it clearly isn't _(at least not on all platforms)_.

But I get your point: monkey-patching `Fixnum` - especially with something that is demonstrably not portable - is even worse.

I'll revisit jimweirich/rake#135 and try to understand why that constant was introduced in the first place, and why it needed that particular name and that particular value... maybe there's no need for it at all, who knows! :wink:

PR #93, opened 2015-12-23
Make Rake::Task already_invoked publicly accessible.

jrafanie:
It's useful to check if a Rake::Task has already been run/invoked, so
let's make this information accessible.

My use case:  I have a rake task which invokes Rails' `environment` rake task with a specific configuration so I want to raise an error if `environment` was previously run/invoked or else my rake task will become a no-op.  I'd rather not have to do this:

```ruby
raise 'xxx' if Rake::Task['environment'].instance_variable_get(:@already_invoked)`
```

### DIFF
--- a/lib/rake/task.rb
+++ b/lib/rake/task.rb
@@ -29,6 +29,10 @@ module Rake
     # location option set).
     attr_reader :locations
 
+    # Has this task already been invoked?  Already invoked tasks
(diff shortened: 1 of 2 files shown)


hsbt:
@jrafanie Thank you. It's reasonable to me. I will release this feature at Rake 11

jrafanie:
Thanks you @hsbt!

Issue #85, opened 2015-11-13
Allow spaces between arguments in Rake task


samzhang111:
`rake my_task[arg1, arg2]`

fails with

`Don't know how to build task 'mytask[arg1, '`

I think this is a pretty big gotcha. It would be nice to support spaces after commas.

drbrain:
As a workaround, use quoting in your shell: `rake "my_task[arg1, arg2]"`

PR #94, opened 2015-12-24
Allow spaces between arguments in Rake tasks

bronzdoc:
This pull request allows input like this <code>rake my_task[arg1, arg2]</code> to be parsed correctly
instead of showing the error <code>Don't know how to build task 'mytask[arg1, '</code> as described
in https://github.com/ruby/rake/issues/85

### DIFF
--- a/lib/rake/application.rb
+++ b/lib/rake/application.rb
@@ -84,6 +84,11 @@ module Rake
       standard_exception_handling do
         @name = app_name
         args = handle_options
+        if args.size > 1 && args.last =~ /.*\]/
+          clean_args = ""
+          args.each{ |arg| clean_args += arg }
+          args = [clean_args]
+        end
         collect_command_line_tasks(args)
       end
     end


hsbt:
Please add testcase.

bronzdoc:
@hsbt there's already a test for this <code>test_can_handle_spaces_between_args</code> in test/test_rake_task_argument_parsing.rb

tarasmatsyk:
hi everyone,
what is the state of this pull request? I see the test in 10.4.2 version but I cannot find the code there.

hsbt:
Why we faced this defects as https://github.com/ruby/rake/issues/85 with `test_can_handle_spaces_between_args`  ? 

I hope to add another test-case.

lib/rake/application.rb
```diff
@@ -84,6 +84,11 @@ def init(app_name='rake')
       standard_exception_handling do
         @name = app_name
         args = handle_options
+        if args.size > 1 && args.last =~ /.*\]/
```
drbrain: The `.*` is not needed in the regular expression.

lib/rake/application.rb
```diff
@@ -84,6 +84,11 @@ def init(app_name='rake')
...
+          clean_args = ""
```
leethomas: This could just be 
```rb
if args.size > 1 && args.last =~ /.*\]/
  args = [args.join]
end
```

right?


Also if the only thing that's stopping this from going out is more tests, I'd be happy to write them.

aycabta:
I think that this is *just shell matter* and you should use quoted string literal. In the Bash example, please check [Bash Reference Manual](https://www.gnu.org/software/bash/manual/bash.html)'s several points, [Quoting](https://www.gnu.org/software/bash/manual/bash.html#Quoting) and [Word Splitting](https://www.gnu.org/software/bash/manual/bash.html#Word-Splitting).

The quoted string literal and word splitting of shell is so complex because of histrical background and Rake is sometimes called by complex shell scripts, such as inside CI, Dockerfile and build scripts. I think this Pull Request is precarious.

For another thing, task name can take any string:

```ruby
task :'task]' do
  puts 'This is task]'
end
```

The task name what is ended of "]" is within the realm of possibility, for example, some kind of model number because Ruby is general-purpos programming language.

rickhull:
@aycabata's concerns make sense to me.  I think @drbrain's [comment here](https://github.com/ruby/rake/issues/85#issuecomment-156537370) is the right approach.  Shell strings, quoting, splitting, etc are gnarly, and I suspect it's best to bite the bullet and be explicit by quoting embedded spaces.

PR #101, opened 2016-01-22
Fix failing tests with JRuby

hsbt:
We have some test failure with JRuby. Rake 11 should be support latest JRuby.

https://travis-ci.org/ruby/rake/builds/103804279


hsbt:
Filed https://github.com/jruby/jruby/issues/3653

PR #74, opened 2015-09-24
Fix special_return method's result wrapping to be more dynamic

tscholz:
Hi,

there is a little issue with the FileList SPECIAL_RETURNs:
```
class CustomFileList < Rake::FileList
end

c = CustomFileList.new #=> []
c.class #=> CustomFileList

c.uniq.class #=> Rake::FileList but should be a CustomFileList
```
What do you think about this fix?

PR #108, opened 2016-02-11
Replace file lists internal calls to `new` by a more generic `self.class.new`

tscholz:
Hi there,

this is a follow up of #74.

Some of file list's methods returning a new FileList instance are not returning the expected type of objects in case of inheritance from FileList. I fixed this in the mentioned PR but forgot to fix this for all methods. I apologise  for the inconvenience.

Please let me know if I should add extra tests for this. Currently I rely on the existing tests.

### DIFF
--- a/lib/rake/file_list.rb
+++ b/lib/rake/file_list.rb
@@ -189,7 +189,7 @@ module Rake
       result = @items * other
       case result
       when Array
-        FileList.new.import(result)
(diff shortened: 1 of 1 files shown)


hsbt:
@tscholz Thanks. It's enough. I merged this.

tscholz:
Thank you!

PR #113, opened 2016-03-07
Correctly handle bad encoding in exception messages

tbrisker:
TraceOutput#trace_on has a bug that causes the matching to fail if a string is incorrectly encoded - for example encoded as 'US-ASCII' but contains unicode characters. This causes a new exception to be thrown, losing the stack trace of the previous exception.
This can occur, for example, in Ruby 1.9 when one of the dependencies of a project includes a file that contains unicode but does not specify an encoding (as the default for Ruby 1.9 is US-ASCII). 
This was the case in a certain version of the Puma web server, which was fixed in https://github.com/puma/puma/pull/924 but was very difficult to debug  due to the incorrect stack trace provided by rake.

### DIFF
--- a/lib/rake/trace_output.rb
+++ b/lib/rake/trace_output.rb
@@ -13,7 +13,7 @@ module Rake
       else
         output = strings.map { |s|
           next if s.nil?
-          s =~ /#{sep}$/ ? s : s + sep
+          s.end_with?(sep) ? s : s + sep
         }.join
       end
       out.print(output)
--- a/test/test_rake_application.rb
+++ b/test/test_rake_application.rb
@@ -1,3 +1,4 @@
+#encoding: UTF-8
 require File.expand_path('../helper', __FILE__)
 
 class TestRakeApplication < Rake::TestCase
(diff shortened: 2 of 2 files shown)


tbrisker:
I believe the test failure is unrelated to my changes.

hsbt:
@tbrisker I update workaround for JRuby at https://github.com/ruby/rake/commit/148374241f28fbb9ddbf76bd9cc7b8fb5912306d . Can you rebase your patch?

tbrisker:
@hsbt I have rebased to latest version. Thank you.

hsbt:
omg, TestRakeFunctional#test_signal_propagation_in_tests is fragile tests with JRuby series. It's enough for your patch. I merge this.

bf4:
fwiw, I wrote a [pretty thorough](https://github.com/rspec/rspec-support/blob/3c86358e132970c96d31e9f6c3aea508b9b4e094/lib/rspec/support/encoded_string.rb) [string-encoding handling class](https://github.com/rspec/rspec-support/blob/3c86358e132970c96d31e9f6c3aea508b9b4e094/spec/rspec/support/encoded_string_spec.rb) for rspec-support (which I [gemified as EncodedString](https://github.com/bf4/encoded_string))

PR #117, opened 2016-03-12
Add possibility to specify test task prerequisites

timsly:
This PR allows to specify dependencies(prerequisites) for test task
```ruby
Rake::TestTask.new do |t|
  t.deps = :setup
end
```

### DIFF
--- a/lib/rake/testtask.rb
+++ b/lib/rake/testtask.rb
@@ -69,6 +69,9 @@ module Rake
     # Description of the test task. (default is 'Run tests')
     attr_accessor :description
 
+    # Task prerequisites.
+    attr_accessor :deps
+
     # Explicitly define the list of test files to be included in a
     # test.  +list+ is expected to be an array of file names (a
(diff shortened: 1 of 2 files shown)


hsbt:
It's enough to use only one dependency? I feels following use-case with `deps` name.

```ruby
Rake::TestTask.new do |t|
  t.deps = [:prepare, :setup]
end
```

Any thought?

timsly:
@hsbt Current implementation supports both `:env` and `[:env1, :env2]` approach. See [here](https://github.com/ruby/rake/pull/117/files#diff-0ebb603bdf4349a31f48ce4e63ed8110R104)

hsbt:
ah, I confirmed `Array()` behavior with array argument.

```
>> Array(:foo)
=> [:foo]
>> Array([:foo, :bar])
=> [:foo, :bar]
```

hsbt:
Thank you for pull request.

PR #118, opened 2016-03-13
Fix verbose under JRuby

pglombardo:
Under JRuby, using `rake/testtask` with the verbose flag broke:
"jruby: unknown option --verbose"

The error in a Travis test run:
https://travis-ci.org/appneta/ruby-traceview/jobs/115183595

`--verbose` isn't a valid command line option for the jruby binary.  It uses `-v`.

This pull request is to use `-v` instead when running under JRuby.


hsbt:
@pglombardo Thank you! I will merge this after appveyor fails fixed. It is not related this changes.

ref. https://github.com/rails-api/active_model_serializers/pull/1585

hsbt:
I released Rake 11.1.1 with https://github.com/ruby/rake/issues/120

PR #120, opened 2016-03-14
Use -W2 instead of --verbose for JRuby

hsbt:
Another patch of https://github.com/ruby/rake/pull/118

### DIFF
--- a/lib/rake/testtask.rb
+++ b/lib/rake/testtask.rb
@@ -128,7 +128,7 @@ module Rake
       opts = @ruby_opts.dup
       opts.unshift("-I\"#{lib_path}\"") unless @libs.empty?
       opts.unshift("-w") if @warning
-      opts.unshift('--verbose') if @verbose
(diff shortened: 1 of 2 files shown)


lib/rake/testtask.rb
```diff
@@ -128,7 +128,7 @@ def ruby_opts_string # :nodoc:
       opts = @ruby_opts.dup
       opts.unshift("-I\"#{lib_path}\"") unless @libs.empty?
       opts.unshift("-w") if @warning
-      opts.unshift('--verbose') if @verbose
+      opts.unshift('-W2') if @verbose
```
bf4: might this conflict with the `-W` option passed to Rake to show task source files?

hsbt: I confirmed `-W` is same as `-W2` on CRuby, JRuby, Rubinius. I changed this. Thanks.

Issue #106, opened 2016-02-04
Description is cut off when running `rake -T` if the description contains dots followed by (white) space


s2k:
Let's assume the following Rakefile:

```Ruby
desc 'Have dots .. in the description'
task :odd do
  puts "doing soething useful..."
end
```

When I run ```rake -T```in the same directory I get this:
```bash
rake -T
rake odd  # Have dots .
```

Notice that the description is cut off after the first dot, even though the description i not particularly long.

However, if there's *no* space after the two dots (```desc 'Have dots ..in the description'```), then the complete description is printed:
```bash
#rake -T
rake odd  # Have dots ..in the description
````
Notes: 

* When using a non-breaking space (ALT-spacebar on OS X) directly after the dots, the complete description is printed as well. This can be used as a workaround.
* It doesn't matter whether or not the dots appear consecutively in the description. Even with ```desc 'Have dots . in . the description'```the text it cut off.

I consider this cutting off of text from the description a bug, since it even happens in relatively short descriptions and other non-dot descriptions of the same length are not cut off either.

Context:
* OS X 10.11.3 (15D21)
* rvm 1.26.11
* Ruby 2.3.0p0 (2015-12-25 revision 53290) [x86_64-darwin15]
* Rake version 10.5.0

hsbt:
@s2k Thank you for your reporting. It seems to be defects.

PR #134, opened 2016-05-15
Update comment parsing rules

leethomas:
Closes https://github.com/ruby/rake/issues/106. Also makes `first_sentence` lookout for exclamation marks. For example:

```
# task name is testing
desc "This drops your database! Make sure you've backed up your db."
```

Running `rake -T` will show
`rake testing            # This drops your database`


hsbt:
Thanks! I have no concerns to your patch.

PR #135, opened 2016-05-16
Don't set if args are blank

bakunyo:
I want to change the action of named arguments #with_defaults.
followings are example.
```
# rake file
task :some_task => [:a, :b] do |task, args|
  args.with_defaults({a: 'a', b: 'b'})
  p args.to_hash
end

# execute(now master)
rake some_task[,c]  # => {:a=>"", :b=>"c"}

# execute(expected)
rake some_task[,c]  # =>  {:a=>"a", :b=>"c"}
```
Rake::TaskArguments are separated by comma, so when I want default value for `:a` but specify for `:b`, want to use arguments `[,any]`.

Please let me know if any problems, I will fix my code and resend pull request.

### DIFF
--- a/lib/rake/task_arguments.rb
+++ b/lib/rake/task_arguments.rb
@@ -17,7 +17,8 @@ module Rake
       @hash = {}
       @values = values
       names.each_with_index { |name, i|
-        @hash[name.to_sym] = values[i] unless values[i].nil?
+        next if values[i].nil? || values[i] == ''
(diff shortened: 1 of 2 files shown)


hsbt:
It's reasonable. I feld `[,c]` means `[nil, c]` too. It's better to use default argument.

bakunyo:
Thank you!

PR #138, opened 2016-06-09
Spawn options for sh

drbrain:
This allows any spawn option to be used with `sh`.

This also improves the printed command output when an ENV hash is supplied.


avdi:
I am excite!

hsbt:
Looks good. I'm going to release this at Rake 11.2.0.

PR #142, opened 2016-06-22
Do not run tasks when their prerequisites fail

ptarjan:
If tasks A and B both depend on C, they race to grab the invocation
lock on C. If A wins, it sets @already_invoked and executes C while
holding the lock. When C completes, A releases the lock, B acquires it
and immediately returns since C is @already_invoked. Unfortunately,
this does not distinguish failed execution of C. Instead, if executing
C raises an exception for A, we want to raise the same exception in B
so it is aborted.

### DIFF
--- a/lib/rake/multi_task.rb
+++ b/lib/rake/multi_task.rb
@@ -4,6 +4,42 @@ module Rake
   # parallel using Ruby threads.
   #
   class MultiTask < Task
+
+    # Same as invoke, but explicitly pass a call chain to detect
+    # circular dependencies. This is largely copied from Rake::Task
+    # but has been updated such that if multiple tasks depend on this
+    # one in parallel, they will all fail if the first execution of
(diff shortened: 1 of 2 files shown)


drbrain:
Can you add a test for this as well?

I think something like this will adequately simulate cross-thread exceptions (but I didn't try it):

```ruby
  def test_cross_thread_prerequisite_failures
    failed = false

    task :fail_once do
      fail_now = !failed
      failed = true
      raise 'failing once' if fail_now
    end

    task a: :fail_once
    task b: :fail_once

    assert_raises RuntimeError do
      Rake::Task[:a].invoke
    end

    assert_raises RuntimeError do
      Rake::Task[:b].invoke
    end
  end
```

ptarjan:
@drbrain added. Thanks for the starting point. I had to make one of them a `multitask` to run my code

PR #144, opened 2016-06-26
Remove trailing slash on a link in README

astorije:
Otherwise:

screen shot 2016-06-26 at 00 20 19

Note the trailing slash not being part of the link nor the label.
I couldn't help myself not to see this 😄


hsbt:
👍 Thank you.

PR #149, opened 2016-07-28
Remove `rake/contrib`

hsbt:
I decided to extract `rake/contrib` files from Rake repository.


hsbt:
created. https://github.com/ruby/rake-contrib

drbrain:
It seems like we should move rubinius back to an allowed failure.

PR #166, opened 2016-09-30
skip tests with JRuby 9k. 

hsbt:
ref. https://travis-ci.org/ruby/rake/jobs/163951069

It caused by `Dir.chdir` specification on JRuby


hsbt:
I need to investigate to behavior with `jruby9?`...

PR #171, opened 2016-10-20
Represent a TaskArgument object such that it's not confused with a hash.

aquateen:
`ta = Rake::TaskArguments.new([:a, :b, :c], [1, 2, 3])`

### Currently
`puts ta # {:a=>1, :b=>2, :c=>3}`

### Proposal
`puts ta # #<Rake::TaskArguments a: 1, b: 2, c: 3>`

### Why?
The current implementation leads one to believe they are working with a hash. This can lead to some confusion if you try doing something like `ta.a = 'one'` which will silently ignore the assignment.

In addition, I think `def_method` is preferable to `method_missing` to create accessors. That way `ta.a = 'one'` would generate a `NoMethodError`. Thoughts?

Shoutout to @eric-hu, with whom I discovered this issue.

### DIFF
--- a/lib/rake/task_arguments.rb
+++ b/lib/rake/task_arguments.rb
@@ -72,11 +72,15 @@ module Rake
     end
 
     def to_s # :nodoc:
-      @hash.inspect
+      inspect
     end
 
     def inspect # :nodoc:
-      to_s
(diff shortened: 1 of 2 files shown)


hsbt:
👍  I prefer this inspection message. Thanks!

PR #174, opened 2016-11-28
Actions adore keyword arguments

JoshCheek:
Consider this task:

```ruby
task :t, [:arg] do |task, arg: 'default_value'|
  emotion = "\e[32m^_^\e[0m"
  emotion = "\e[31mt.t\e[0m" if ARGV[0] =~ /t\[(.*)\]$/ && $1 != arg
  puts "task #{task.name.inspect} got #{arg.inspect} #{emotion}"
end
```

Run this on your machine right now and you'll see the task is in tears. But this commit turns tearful eyes into smiling ones, observe:

```sh
$ rake t
task "t" got "default_value" ^_^
$ rake t[custom_value]
task "t" got "default_value" t.t

$ ruby -Ilib exe/rake t
task "t" got "default_value" ^_^
$ ruby -Ilib exe/rake t[custom_value]
task "t" got "custom_value" ^_^
```

It accomplishes this by always invoking the action in the same way. [Previously](https://github.com/ruby/rake/blob/59856100815b841624269f817c815f54921bf321/lib/rake/task.rb#L251-L256), Rake invoke the action with different argument structures based the arity.

```ruby
case act.arity
when 1
  act.call(self)
else
  act.call(self, args)
end
```

I assume it's expecting one of these:

```ruby
proc { |task| }.arity           # => 1
proc { |task, options| }.arity  # => 2
```

However this leads numerous task signatures to be invoked incorrectly. Eg, consider the case of our tearful task above, here is why it cries.

```ruby
proc { |task, arg: 'default_value'| }.arity # => 1
```

The solution is nice and easy: always invoke the block the same way.

Why was this ever a thing at all, then? Looks like it was added in 2007 about a month after Ruby 1.8.6 was released. https://github.com/ruby/rake/commit/925fbb80e890029ac9310bc60270890699efc073 Probably something was just different back then, eg the ([difference](https://github.com/ruby/ruby/blob/9b383bd/eval.c#L8356-L8415)) between lambdas and procs was whether you did `prc.call` or `prc.yield`

### DIFF
--- a/test/test_rake_task_with_arguments.rb
+++ b/test/test_rake_task_with_arguments.rb
@@ -80,6 +80,31 @@ class TestRakeTaskWithArguments < Rake::TestCase
     assert_equal [:a, :b, :c, :d], notes
   end
 
+
+  def test_actions_adore_keywords
+    notes = []
+    t = task :t, [:reqr, :ovrd, :dflt] # required, overridden-optional, default-optional
+    verify = lambda do |name, expecteds, actuals|
+      notes << name
+      assert_equal expecteds.length, actuals.length
+      expecteds.zip(actuals) { |e, a| assert_equal e, a, "(TEST #{name})" }
+    end
+
+    t.enhance { |dflt: 'd', **| verify.call :a, ['d'], [dflt] }
+    t.enhance { |ovrd: '-', **| verify.call :b, ['o'], [ovrd] }
+    t.enhance { |reqr:    , **| verify.call :c, ['r'], [reqr] }
+
+    t.enhance { |t2, dflt: 'd', **| verify.call :d, [t,'d'], [t2,dflt] }
+    t.enhance { |t2, ovrd: 'd', **| verify.call :e, [t,'o'], [t2,ovrd] }
+    t.enhance { |t2, reqr:    , **| verify.call :f, [t,'r'], [t2,reqr] }
+
+    t.enhance { |t2, dflt: 'd', reqr:, **| verify.call :g, [t,'d','r'], [t2,dflt,reqr] }
+    t.enhance { |t2, ovrd: '-', reqr:, **| verify.call :h, [t,'o','r'], [t2,ovrd,reqr] }
+
+    t.invoke('r', 'o')
+    assert_equal [*:a..:h], notes
+  end
+
   def test_arguments_are_passed_to_block
     t = task(:t, :a, :b) { |tt, args|
       assert_equal({ a: 1, b: 2 }, args.to_hash)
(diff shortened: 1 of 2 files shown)


test/test_rake_task_with_arguments.rb
```diff
@@ -82,6 +82,10 @@ def test_actions_of_various_arity_are_ok_with_args
 
 
   def test_actions_adore_keywords
+    # A brutish trick to avoid parsing. Remove it once support for 1.9 and 2.0 is dropped
+    # https://ci.appveyor.com/project/ruby/rake/build/1.0.301
+    skip 'Keywords aren\'t a feature in this version' if RUBY_VERSION =~ /^1|^2\.0/
+    eval <<-RUBY, binding, __FILE__, __LINE__
```
nobu: You need `+1` to `__LINE__`

nobu:
A failure in jruby-head seems a bug of jruby.

n054:
@JoshCheek let me see if I can help with Travis intigration. @n054

JoshCheek:
@n054 I'd appreciate it ^_^

n054:
@JoshCheek investigating the Travis test shows test#482.1-482.7 are passing.
Test#482.8&482.9 show TestRakeTaskWithArguments#test_actions_adore_keywords:
ArgumentError:missing keyword:  reqr

JoshCheek:
Looks like it's a JRuby bug. Surprising, maybe I just use keywords a lot more than other people?

Here's the program:

```sh
$ cat f.rb
p proc { |a, b:| [a, b] }.call("a-val", b: "b-val")
```

MRI handles it.

```sh
$ chruby 2.3.2
$ ruby -v
ruby 2.3.2p217 (2016-11-15 revision 56796) [x86_64-darwin15]
$ ruby f.rb
["a-val", "b-val"]
```

Rbx handles it

```sh
$ chruby rbx
$ rbx -v
rubinius 3.60 (2.3.1 5d5e49f7 2016-09-21 3.9.0) [x86_64-darwin15.6.0]
$ rbx f.rb
["a-val", "b-val"]
```

JRuby blows up.

```sh
$ chruby jruby
$ jruby -v
jruby 9.1.5.0 (2.3.1) 2016-09-07 036ce39 Java HotSpot(TM) 64-Bit Server VM 25.71-b15 on 1.8.0_71-b15 +jit [darwin-x86_64]
$ jruby f.rb
ArgumentError: missing keyword: b
  block in f.rb at f.rb:-1
         <main> at f.rb:1
```

JoshCheek:
Looks @nobu beat me to reporting: https://github.com/jruby/jruby/issues/4344

PR #182, opened 2016-12-23
When pattern is nil, enhance_with_matching_rule will not fail

lzap:
This fixes the

    NoMethodError: undefined method `match' for nil:NilClass

error with gettext/gettext_i18n_rails.

https://github.com/ManageIQ/manageiq/issues/4774


lzap:
I put an `inspect` and it's full of nils strangely: `[[nil, [], [], nil], [nil, [], [], nil], [nil, [], [], nil], [nil, [], [], nil], [nil, [], [], nil], [nil, [], [], nil]]`

PR #183, opened 2016-12-28
FileTask#out_of_date? calls depended FileTask#out_of_date? for execution

### LINKED ISSUE #92: Rake -n does not mark dependents for Execution
I have a Rakefile with these file tasks:

``` ruby
file "fileA" => %W[] do |t| sh %Q[echo contentA >#{t.name}] end
file "fileB" => %W[fileA] do |t| sh %Q[(cat fileA; echo transformationB) >#{t.name}] end
file "fileC" => %W[fileB] do |t| sh %Q[(cat fileB; echo transformationC) >#{t.name}] end
```

I perform this setup, where everything is built and then the first file in the chain (`fileA`) is updated:

``` sh
$ rake fileC
echo contentA >fileA
(cat fileA; echo transformationB) >fileB
(cat fileB; echo transformationC) >fileC
$ touch fileA
```

And now `rake -n` tells me that only `fileB` needs to be updated...

``` sh
$ rake -n fileC
** Invoke fileC (first_time, not_needed)
** Invoke fileB (first_time)
** Invoke fileA (first_time, not_needed)
** Execute (dry run) fileB
```

Even though it knows it must update both `fileB` and `fileC`:

``` ruby
$ rake fileC
(cat fileA; echo transformationB) >fileB
(cat fileB; echo transformationC) >fileC
```

I expected `rake -n` to tell me this:

``` ruby
$ rake -n fileC
[...]
** Execute (dry run) fileB
** Execute (dry run) fileC
```

The problem seems to stem from the fact that the internal function `out_of_date?` (in `file_task.rb`) does not take into account when a file _should_ have updated its timestamp (but didn't, because this is a dry-run).

aycabta:
This fixes #92 what says below:

> The problem seems to stem from the fact that the internal function `out_of_date?` (in `file_task.rb`) does not take into account when a file should have updated its timestamp (but didn't, because this is a dry-run).

I think so. In this Pull Request, `FileTask#out_of_date?` calls dependency `FileTask`'s `#out_of_date?` via `FileTask#needed?`.

### DIFF
--- a/lib/rake/file_task.rb
+++ b/lib/rake/file_task.rb
@@ -29,7 +29,14 @@ module Rake
 
     # Are there any prerequisites with a later time than the given time stamp?
     def out_of_date?(stamp)
-      @prerequisites.any? { |n| application[n, @scope].timestamp > stamp }
+      @prerequisites.any? { |p|
+        ptask = application[p, @scope]
+        if ptask.instance_of?(Rake::FileTask)
+          ptask.timestamp > stamp || ptask.needed?
+        else
+          ptask.timestamp > stamp
+        end
+      }
     end
 
     # ----------------------------------------------------------------
--- a/lib/rake/file_task.rb
+++ b/lib/rake/file_task.rb
@@ -29,12 +29,12 @@ module Rake
 
     # Are there any prerequisites with a later time than the given time stamp?
     def out_of_date?(stamp)
-      @prerequisites.any? { |p|
-        ptask = application[p, @scope]
-        if ptask.instance_of?(Rake::FileTask)
-          ptask.timestamp > stamp || ptask.needed?
+      @prerequisites.any? { |prereq|
+        prereq_task = application[prereq, @scope]
+        if prereq_task.instance_of?(Rake::FileTask)
+          prereq_task.timestamp > stamp || prereq_task.needed?
         else
-          ptask.timestamp > stamp
+          prereq_task.timestamp > stamp
         end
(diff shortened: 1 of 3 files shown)


lib/rake/file_task.rb
```diff
@@ -29,7 +29,14 @@ def timestamp
 
     # Are there any prerequisites with a later time than the given time stamp?
     def out_of_date?(stamp)
-      @prerequisites.any? { |n| application[n, @scope].timestamp > stamp }
+      @prerequisites.any? { |p|
```
aycabta: Oh, I made a silly mistake. It's as you says. I added a commit 679ccb8.

test/test_rake_functional.rb
```diff
@@ -295,6 +295,26 @@ def test_rules_chaining_to_file_task
...
+    }
+  end
+
+  def test_update_midway_through_chaining_to_file_task
+    rakefile_file_chains
+
+    rake "-n"
+    assert_equal(%w{fileA fileB fileC default}, dryrun_tasks)
+    rake
+    sleep 1 # for stride seconds surely for timestamp
```
olleolleolle: Just a passer-by reading this comment. I'm confused by it. Can you reword it, to make it less ambiguous?

What I believe it wants to say: Ensure the timestamp is on a new second.

aycabta: Thanks, I added a commit to fix the comment!

PR #184, opened 2016-12-29
Travis: add 2.4.0 to matrix, update patch versions

olleolleolle:
This PR adds 2.4.0 as a test target in Travis.

It also takes the patch versions of other rubies from the `ruby-install --latest` output and updates them.

PR #185, opened 2016-12-29
The "raise" with no arguments raises the exception in $!, nested case too

aycabta:
Ruby 2.4 and later, `raise`'s behavior has been changed. Nested `raise` with an argument raises only the argument, it doesn't have outer exception. So I removed the argument. The `raise` with no arguments raises what contains nested exceptions.

This fixes a failure of `TestRakeApplication#test_display_exception_details_cause_loop` that @hsbt refers at https://github.com/ruby/rake/pull/184. This [passed all Ruby versions in Travis CI](https://travis-ci.org/aycabta/rake/builds/187525488).

...And this will fix Travis CI failure of https://github.com/ruby/rake/pull/183.

### DIFF
--- a/test/test_rake_application.rb
+++ b/test/test_rake_application.rb
@@ -80,11 +80,11 @@ class TestRakeApplication < Rake::TestCase
     begin
       begin
         raise "cause a"
-      rescue => a
+      rescue
         begin
           raise "cause b"
         rescue
-          raise a
+          raise
         end
       end
     rescue => ex


nobu:
This just re-raises "cause_b" exception, and makes no sense.

aycabta:
> This just re-raises "cause_b" exception

I think so, but I didn't want to say so.

I do the complementary explanation that I did. The "nested exception" means [chaining `Exception` with `Exceptin#cause`](https://github.com/ruby/rake/blob/93e55a4ef1dbaee42f0f355f86d837c4e2551fc1/lib/rake/application.rb#L213). The test checks just it.

aycabta:
https://twitter.com/hsbt/status/816458882541228032 (Japanese, I read with translation)

I grasped the overview around this tweet and its reply tree. If that tells you anything, I just made this Pull Request for 2.4's behavior, and I didn't understand the purpose of the test. I will leave it to appropriate one's judgement.

ref. The test is added for https://github.com/jimweirich/rake/issues/272 at first.

nobu:
With this change, the test is almost same as `test_display_exception_details_cause`, except for an extra re-raising, and no meaning of `loop`.

aycabta:
OK, I got the picture. I'm thinking what `test_display_exception_details_cause_loop` should be deleted.

PR #186, opened 2017-01-04
Remove an extra assertion

nobu:
As the result of https://bugs.ruby-lang.org/issues/13043, now `Exception#cause` should not have a loop.

In the example https://github.com/jimweirich/rake/issues/272, the code doesn't seem to intend the loop itself but just re-raising the first exception instead of the next exception.  Therefore I consider the last assertion superfluous.


hsbt:
Make sense. Thank you.

Issue #201, opened 2017-05-02
FileUtils#sh occurs TypeError when using hash style options.


igaiga:
Until rake v.11.1.2, we can run sh commands with hash style option, like the follow. But we have "TypeError" now.

```
task :dir do
  sh "pwd", {chdir: "/home"}, { vernbose: true }
end
# before rake v.11.1.2 #=> /home [OK]
# after rake v.11.2.0  #=> TypeError: no implicit conversion of Hash into String [Error]
```

On the other hand, to write codes same as previous one like the follow, it doesn't work before v.11.1.2.

```
task :dir do
  verbose(true) do
    sh "pwd", {chdir: "/home"}
  end
end
# before rake v.11.1.2 #=> ArgumentError: no such option: chdir [Error]
# after rake v.11.2.0  #=> /home [OK]
```

Is it spec or bug?

### information : About the "TypeError: no implicit conversion of Hash into String"

```
task :dir do
  sh "pwd", {chdir: "/home"}, { vernbose: true }
end
# before rake v.11.1.2 #=> /home
# after rake v.11.2.0  #=> TypeError: no implicit conversion of Hash into String
```

We have the "TypeError: no implicit conversion of Hash into String" error from v.11.2.0.

The error occurs on the lib/rake/file_utils.rb:56 line.
It looks like to become occur from this code changes.
https://github.com/ruby/rake/pull/138/commits/226b8a0dbfca92dfb4202dcf589b53400584d6f3#diff-f7fcb87d482d055b9df1e86fd76ae06fR56

The code use Kernel#system method. After the change, arguments for system method is increasing.

v.11.1.2 
```
system('pwd',  {chdir: "/home"})
#=> /home
```

v.11.2.0
```
system('pwd',  {chdir: "/home"}, {})
#=> TypeError: no implicit conversion of Hash into String
```

PR #202, opened 2017-05-02
Change FileUtils#sh to adopt with hash style option.

igaiga:
Fix https://github.com/ruby/rake/issues/201 problem if that is a bug.


hsbt:
It's not a specification. 

Thank you for your patch.

PR #203, opened 2017-05-20
Phony tasks

### LINKED ISSUE #25: Marking tasks as :phony fails to prevent rebuild
The [:phony task](https://github.com/ruby/rake/blob/99ca610e3d166991e110bdf6d1c52bfdd8aae5ff/lib/rake/phony.rb) fails to prevent rebuild of FileTasks dependent on non-file tasks.

### [lib/rake/phony.rb](https://github.com/ruby/rake/blob/99ca610e3d166991e110bdf6d1c52bfdd8aae5ff/lib/rake/phony.rb)

```ruby
# Defines a :phony task that you can use as a dependency. This allows
# file-based tasks to use non-file-based tasks as prerequisites
# without forcing them to rebuild.
#
# See FileTask#out_of_date? and Task#timestamp for more info.

require 'rake'

task :phony

Rake::Task[:phony].tap do |task|
  def task.timestamp # :nodoc:
    Time.at 0
  end
end
```

### Code reproducing bug

```ruby
require 'rake/phony'

task :default => 'final.file'

file 'final.file' => ['source.file', :intermediate] do |f|
  touch f.name
end

# Marked as phony, so should not force FileTask to rebuild
task :intermediate => [:phony] do |t|
  puts "#{t.name} task"
end
```

### Bug demo

```
$ touch source.file      # Create the source file
$ rake                   # Run Rake to build 'final.file'
intermediate task
touch final.file
$ rake                   # Rake should NOT rebuild, but does.     <<<<<<<<
intermediate task
touch final.file
```
---
Mentioned [here](http://stackoverflow.com/questions/28535829)

juank-pa:
Fixes #25 
While reporting non-file timestamps as always `Time.now` is definitively the best approach to make tasks more predictable, this left the phony task feature broken.

This could be solved by instead add the ability for tasks to get marked as phony so that they report the expected time directly to tasks depending on it.


test/test_rake_task.rb
```diff
@@ -296,24 +296,26 @@ def test_all_prerequisite_tasks_includes_self_on_cyclic_dependencies
     assert_equal [a, b], a.all_prerequisite_tasks.sort_by { |t| t.name }
   end
 
-  def test_timestamp_returns_now_if_all_prereqs_have_no_times
+  def test_timestamp_returns_now_regardless_prereq_times
```
juank-pa: Previous test name was misleading because current behavior is to have time stamp always report as now.

test/test_rake_task.rb
```diff
@@ -296,24 +296,26 @@ def test_all_prerequisite_tasks_includes_self_on_cyclic_dependencies
...
     a = task a: ["b", "c"]
-    task :b
-    task :c
+    b = task :b
+    c = task :c
+
+    def b.timestamp() Time.now + 10 end
```
juank-pa: I change the dependencies timestamp here to ensure they do not affect the dependent task timestamp.

mill1000:
This PR was closed without merging the changes. Phony still seems to be broken? Is there another way around this?

mill1000:
Here's a snippet to extend the Task class as a work around

```ruby
# Extend Rake::Task to support phony
module Rake
  class Task
    @phony = false
    def phony
      @phony = true
      self
    end

    def timestamp
      @phony ? Time.at(0) : Time.now
    end
  end
end
```

juank-pa:
Yeah this was still an issue. I can use the workaround. Not a big fan of monkey patching though.

PR #211, opened 2017-06-29
Make rake easier to use as a library

drbrain:
I want to be able to load a set of rake tasks from within another application.  Doing this presently is difficult, and looks something like this:

```ruby
Dir.chdir 'other/dir'

rake = Rake::Application.new
Rake.application = rake

rake.options.trace = true
rake.options.backtrace = true
rake.options.rakelib = %w[rakelib]

rake.load_rakefile

puts rake.tasks
```

Figuring out which options you need to set is hard (and not documented), figuring out you need to set `Rake.application` is not documented, and the process of getting your Rakefile loaded is not documented.

This PR replaces the above with a much simpler option:

```ruby
Dir.chdir 'other/dir'

rake = Rake.with_application do |rake|
  rake.load_rakefile
end

puts rake.tasks
```

Figuring out which options is now hidden from the user.

There are a few other API niceties in here like being able to create a new Rake application using a separate set of command line arguments from `ARGV` (so you don't have to overwrite `ARGV`).

Some tests that use `capture_io` got a little more confusing because the output IO was set to real-`$stderr` before `capture_io` changed `$stderr` to a StringIO.

### DIFF
--- a/lib/rake/application.rb
+++ b/lib/rake/application.rb
@@ -74,19 +74,19 @@ module Rake
     # If you wish to build a custom rake command, you should call
     # +init+ on your application.  Then define any tasks.  Finally,
     # call +top_level+ to run your top level tasks.
-    def run
+    def run(argv = ARGV)
       standard_exception_handling do
-        init
+        init 'rake', argv
         load_rakefile
         top_level
       end
     end
 
     # Initialize the command line parameters and app name.
-    def init(app_name="rake")
+    def init(app_name="rake", argv = ARGV)
       standard_exception_handling do
         @name = app_name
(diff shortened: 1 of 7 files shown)


lib/rake/application.rb
```diff
@@ -74,19 +76,19 @@ def initialize
     # If you wish to build a custom rake command, you should call
     # +init+ on your application.  Then define any tasks.  Finally,
     # call +top_level+ to run your top level tasks.
-    def run
+    def run(argv = ARGV)
```
Eric-Guo: Seems it will caused #233 problem

PR #213, opened 2017-07-05
Make file task faster by not having to hit disk multiple times

shreyasbharath:


drbrain:
Can you provide a before and after benchmark?

drbrain:
… specifically, I'd like to know if trading `stat(2)` for creation and garbage collection of an `Errno::ENOENT` and its backtrace gives an overall speed benefit.

shreyasbharath:
I can provide benchmarks, any suggestions on how best to measure it?

I used the same trick in a different project I was working on and I noticed some decent speed improvements when dealing with thousands of files. I didn't benchmark this change in isolation however.

lib/rake/file_task.rb
```diff
@@ -19,10 +19,10 @@ def needed?
 
     # Time stamp for file task.
     def timestamp
-      if File.exist?(name)
-        File.mtime(name.to_s)
-      else
-        Rake::LATE
+      begin
```
shreyasbharath: We can possibly get rid of the ! File.exist?(name) in the needed? function above too.

lib/rake/file_task.rb
```diff
@@ -19,10 +19,10 @@ def needed?
...
+        return File.mtime(name.to_s)
```
mblumtritt: The `return` statement is not required here.

lib/rake/file_task.rb
```diff
@@ -19,10 +19,10 @@ def needed?
...
+      rescue Errno::ENOENT
+        return Rake::LATE
```
mblumtritt: dito: The `return` statement is not required here.

shreyasbharath:
Okay here's a simple rakefile I used to benchmark the change -

    require 'rubygems'
    require 'rake'

    prerequisites = FileList.new('test/*').to_a

    file 'foo.txt' => prerequisites do
      touch 'foo.txt'
    end

Prior to running the rake file task, I created 10,000 test files like so -
    0.upto(9999) { |i| FileUtils.touch(i.to_s) }

Here are my benchmarks

Pre-change
pre-change

Post-change
post-change

lib/rake/file_task.rb
shreyasbharath: Fixed.

lib/rake/file_task.rb
shreyasbharath: Fixed.

hsbt:
I'm not sure to improve this change for All of the platforms contained Linux/macOS.

shreyasbharath:
@hsbt I can run the same benchmark on Linux but I don't have a Macbook sorry.

yuki24:
@shreyasbharath it would be great if you could provide us with a benchmark script using the [Benchmark](https://ruby-doc.org/stdlib-2.4.2/libdoc/benchmark/rdoc/Benchmark.html) or [the benchmark-ips gem](https://github.com/evanphx/benchmark-ips). This will allow us to easily run your benchmark script on different platforms. I'm also curious to know why you'd need thousands of file tasks since it seems unusual to me to have to deal with thousands of files tasks.

rickhull:
Here is a bench script with output and numbers on a single CPU Vagrant VM: https://gist.github.com/rickhull/90c7ec5b1abac13e461350595f522d4b

T and F indicate whether the file exists or not.

```
known existing files: 128
Warming up --------------------------------------
       File.exist? T   362.000  i/100ms
       File.exist? F     1.119k i/100ms
     File.exist? T+F   271.000  i/100ms
            rescue T   601.000  i/100ms
            rescue F   141.000  i/100ms
          rescue T+F   112.000  i/100ms
Calculating -------------------------------------
       File.exist? T      3.638k (± 2.0%) i/s -     18.462k in   5.076247s
       File.exist? F     11.523k (± 0.9%) i/s -     58.188k in   5.050013s
     File.exist? T+F      2.741k (± 1.0%) i/s -     13.821k in   5.043371s
            rescue T      6.111k (± 1.0%) i/s -     30.651k in   5.015826s
            rescue F      1.445k (± 1.0%) i/s -      7.332k in   5.075331s
          rescue T+F      1.133k (± 1.0%) i/s -      5.712k in   5.043949s

Comparison:
       File.exist? F:    11523.2 i/s
            rescue T:     6111.4 i/s - 1.89x  slower
       File.exist? T:     3638.5 i/s - 3.17x  slower
     File.exist? T+F:     2740.7 i/s - 4.20x  slower
            rescue F:     1444.8 i/s - 7.98x  slower
          rescue T+F:     1132.6 i/s - 10.17x  slower
```

shreyasbharath:
@rickhull Thanks for posting the benchmark.

Perhaps my naivety got the better of me here as I was optimising for my particular use case only. 8x slower for non-existent is not acceptable.

@hsbt @yuki24 Please feel free to close this.

PR #224, opened 2017-09-19
fix quadratic performance in FileTask#out_of_date?

doudou:
FileTask#out_of_date? has been changed in 462e403a to call #needed?
which calls #out_of_date? recursively. In some cases, where the
graph of dependencies is fairly dense, this leads to quadratic
performance when it was linear before.

Use #all_prerequisite_tasks to avoid the problem. This also saves
a File.exist? test as #timestamp already takes it into account.

### DIFF
--- a/lib/rake/file_task.rb
+++ b/lib/rake/file_task.rb
@@ -30,10 +30,10 @@ module Rake
 
     # Are there any prerequisites with a later time than the given time stamp?
     def out_of_date?(stamp)
-      @prerequisites.any? { |prereq|
+      all_prerequisite_tasks.any? { |prereq|
         prereq_task = application[prereq, @scope]
         if prereq_task.instance_of?(Rake::FileTask)
-          prereq_task.timestamp > stamp || prereq_task.needed?
+          prereq_task.timestamp > stamp || @application.options.build_all
         else
           prereq_task.timestamp > stamp
         end


hsbt:
Thanks. Can you show the benchmark results?

doudou:
> Thanks. Can you show the benchmark results?

I don't have them, it was reported by one of my coworker.

On our system, it went from our rake-based library doing 100% CPU for minutes to 2-3% CPU (the rest is taken by subcommands). It's so dramatic that I didn't see the need to isolate a benchmark.

If having a benchmark is a prerequisite, I'll try to take the time to do it.

aycabta:
I'm original writer of the `#out_of_date?` logic of now at https://github.com/ruby/rake/pull/183. I understand this Pull Request. I think that this is smart arrangement and appropriate for the logic, I gues that it's enough reason to merge this.

However I agreed that benchmark is important for this. In case you didn't know yet, RDoc [uses `NOBENCHMARK` env to enable benchmark test](https://github.com/ruby/rdoc/blob/5f11ddf9b7771bda393fd31a726c518e4503efe9/lib/rdoc/test_case.rb#L9) because CI server sometimes has a bumpy ride performance in the short term, if you want to add benchmark test.

doudou:
Fair enough ... how should I provide this benchmark ? Rake doesn't seem to have a benchmark running facility as @aycabta mentions for RDoc. Just in this PR ?

doudou:
I've added the benchmark code and the results to [the commit message](https://github.com/ruby/rake/pull/224/commits/85fa80dc639f68de2b921a605bc4a35a937099b6)

aycabta:
OK, RDoc [uses the benchmark test at `test/test_rdoc_context.rb`](https://github.com/ruby/rdoc/blob/f4c498dd084d9ec493ffaa75005850956ab591fa/test/test_rdoc_context.rb#L413) and you can see [documents for the assersions of MiniTest](http://www.rubydoc.info/github/seattlerb/minitest/Minitest/Benchmark#instance_method_details).

On the other hand, if you want to just take benchmark result as distinct from test, I think that [benchmark-ips](https://rubygems.org/gems/benchmark-ips) is best choice for it. It's standard for presentation of Rails performance with patch.

aycabta:
Oh, I was writing my comment but later than you.

doudou:
Ping

hsbt:
I heard details of this issue from @aycabta . I'm ok to merge this.

Issue #225, opened 2017-09-22
FileList is not unique


BertramScharpf:
`Rakefile`:

    task :list do
      FileList[ "texput.*", "*.log"].each { |f| puts "==== #{f}" }
    end
    task :clean do
      FileList[ "texput.*", "*.log"].each { |f| rm f }
    end

`ls` output:

    Rakefile
    dummy.log
    texput.log
    texput.tex

Rake commands:

    $ rake list
    ==== texput.log
    ==== texput.texrm texput.log
    ==== dummy.log
    ==== texput.log
    $ rake clean
    rm texput.tex
    rm dummy.log
    rm texput.log
    rake aborted!
    Errno::ENOENT: No such file or directory @ unlink_internal - texput.log
    ...

As you can see, "texput.log" appears twice what yields an error when you call a plain remove.

I already reported this 2.5 years ago and the only thing that happend was that I was critizied because my patch broke the unit test. I refused to fix the unit test as long as the unit test doesn't catch the actual bug. I wanted to report but not to be drawn into the quagmire of questionable code. That dispute seemed to be more important to the maintainers than the misbehaviour of the actual module. Maybe the old narrow-minded maintainers have resigned in the meantime.

drbrain:
You must be referring to #30

Can you update the tests to match the behavior you expect?

It is a common thing to update the tests when you find a bug so the bug does not re-appear.  Typically this burden falls upon the submitter, as they are familiar with and care a lot about their bug, not with the maintainers who can have many other projects to maintain.

drbrain:
The built-in clean task doesn't have this issue because it also [handles races](https://github.com/ruby/rake/blob/f7b96895832360f6a254b67c1b07d0b205494363/lib/rake/clean.rb#L36) in deleting files by checking if the file is already deleted when deletion fails.

Also, the test failures from #30 show there is are `uniq` and `uniq!` methods on Rake::FileList.  This behavior is at least 14 years old, and you have thought most about it recently, so we would like to defer to your expertise in fixing any test changes that are needed.

styd:
Hi @BertramScharpf, I think the intended flow is to call on `existing` first before you iterate on them. But even then, the current implementation still shows duplicate file on the result.

That's why I propose #231.

If that's merged. You can iterate on them like this:
```ruby
task :list do
  FileList[ "texput.*", "*.log"].existing.each { |f| puts "==== #{f}" }
end
task :clean do
[...]

PR #231, opened 2017-10-15
Account for a file that match 2 or more patterns.

styd:
I believe calling `existing` on a FileList should return a unique list because files in a directory are unique. This could also solve issue #30 and #225 if he calls each on the existing FileList.


hsbt:
It's reasonable. Thanks!

PR #235, opened 2017-11-06
make clear that scope is ignored purposely in FileTask.scope_name

rickhull:
I'm not 100% certain on whether this change is desirable or self-evidently so.  But I definitely had to do a double-take when I looked at the method definition and saw that only the last argument was even considered.  The existing comments somewhat explain it.  I think the leading underscore on `_scope` makes the code more self-evident.


rickhull:
Closed and resubmitted as #237  so that I can use my fork's `master` branch for tracking.

PR #243, opened 2017-12-19
Add a Rake.application.running? predicate

leonid-shevtsov:
I have added a `Rake.application.running?` method, that returns true if Rake is running currently, and false otherwise. It is most useful to fence off code that is only needed for Rake tasks, or conversely, not needed. 

This is a common need in Rails apps and the most popular way to see if the code is running within a rake task is checking the command line argument (see SO answers at the bottom). Which might work, but it is messy, and leads to copy-paste of the `File.basename($0) == 'rake'` idiom.  Our project uses that code snippet in three distinct places, but instead of fixing this on a project level, in my opinion, this is a capability that Rake should have out of the box.

StackOverflow answers that suggest checking the script filename:

* https://stackoverflow.com/a/42619194/6678
* https://stackoverflow.com/a/15767148/6678

### DIFF
--- a/lib/rake/application.rb
+++ b/lib/rake/application.rb
@@ -56,6 +56,7 @@ module Rake
       @loaders = {}
       @default_loader = Rake::DefaultLoader.new
       @original_dir = Dir.pwd
+      @running = false
       @top_level_tasks = []
       add_loader("rb", DefaultLoader.new)
       add_loader("rf", DefaultLoader.new)
@@ -77,11 +78,13 @@ module Rake
     # +init+ on your application.  Then define any tasks.  Finally,
     # call +top_level+ to run your top level tasks.
     def run(argv = ARGV)
+      @running = true
       standard_exception_handling do
         init "rake", argv
         load_rakefile
         top_level
       end
+      @running = false
     end
 
     # Initialize the command line parameters and app name.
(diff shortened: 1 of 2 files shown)


yuki24:
> It is most useful to fence off code that is only needed for Rake tasks, or conversely, should not run inside a Rake task.

Interesting, I haven't had the need like this. Could you elaborate on this?

lib/rake/application.rb
```diff
@@ -77,11 +78,13 @@ def initialize
     # +init+ on your application.  Then define any tasks.  Finally,
     # call +top_level+ to run your top level tasks.
     def run(argv = ARGV)
+      @running = true
       standard_exception_handling do
         init "rake", argv
         load_rakefile
         top_level
       end
+      @running = false
```
yuki24: This should be wrapped by an `ensure` block:

```ruby
def run(...)
  ..
ensure
  @running = false
end
```

test/test_rake_application.rb
```diff
@@ -634,4 +634,22 @@ def loader.make_dummy
...
+    end
+
+    rakefile_default
+
+    assert !@app.running?
+    @app.run %w[--rakelib=""]
+    assert !@app.running?
+    assert was_running
+  end
+
```
yuki24: Could you also add a test case where the task execution fails but it correctly sets `@running` back to `false` after failure?

leonid-shevtsov:
Done - sorry for the huge delay.

To elaborate on this:

>It is most useful to fence off code that is only needed for Rake tasks, or conversely, should not run inside a Rake task.

to give one example, we mark changes made from a rake task for audit:

```
if File.basename($PROGRAM_NAME) == "rake"
  PaperTrail.whodunnit = "#{`whoami`.strip}: rake #{ARGV.join ' '}"
end
```

Another places I can recall: 

- extended logging inside Rake tasks - logs that would be spammy in the server process.
- do not start project-specific subsystems that are known to make Rake tasks slow and are not necessary in Rake tasks.

(All of the above assumed a webapp context.)

PR #254, opened 2018-02-21
Removes duplicated inclusion of Rake::DSL

grzuy:
`Rake::DSL` is already included in `Rake::TestCase`

### DIFF
--- a/test/test_rake_file_creation_task.rb
+++ b/test/test_rake_file_creation_task.rb
@@ -4,7 +4,6 @@ require "fileutils"
 
 class TestRakeFileCreationTask < Rake::TestCase
   include Rake
-  include Rake::DSL
 
   DUMMY_DIR = "dummy_dir"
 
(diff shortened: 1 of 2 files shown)


grzuy:
Apprently test build failed only for ruby 2.1, because `bundle install` command was installing rdoc 5.x previously for all builds, but installed rdoc 6.x for the first time with this PR (despite rdoc 6.x released 1 2 months ago), which is no longer compatible with ruby 2.1

grzuy:
Reported issue here https://github.com/ruby/rake/issues/255

hsbt:
@gazay Can you rebase from master?

grzuy:
@hsbt Rebased

yuki24:
I'm merging this PR since this seems like a safe change. Thanks @grzuy!

grzuy:
Thank you!

PR #262, opened 2018-03-20
support did_you_mean >= v1.2.0 which has a breaking change on formatters

gfx:
did_you_mean  v1.2.0 has a breaking change to its formatter api:

https://github.com/yuki24/did_you_mean/compare/v1.1.2...v1.2.0

This PR uses the new `DidYouMean.formatter`, keeping backward compatibility for did_you_mean pre-1.2.0.


hsbt:
Good catch. Thanks!

PR #265, opened 2018-05-24
Fix JRuby detection on JRuby 9.2 in cpu_counter.rb

jeremyevans:
`Java::Java` is no longer defined by default, it's not defined until it is accessed:

```
$ jruby -e 'p defined?(Java::Java)'
nil
$ jruby -e 'p Java::Java'
Java::Java
```

In earlier JRuby versions, `defined?(Java::Java)` returned `"constant"`.

While this doesn't cause an test failures, it does result in the following warning when run on non-Windows systems: `!!!! Missing jruby-win32ole gem: jruby -S gem install jruby-win32ole`, because the first fallback code is to try to load `win32ole`, and on JRuby that prints the warning.

### DIFF
--- a/lib/rake/cpu_counter.rb
+++ b/lib/rake/cpu_counter.rb
@@ -32,7 +32,7 @@ unless Rake::CpuCounter.method_defined?(:count)
     require 'rbconfig'
 
     def count
-      if defined?(Java::Java)
+      if defined?(JRUBY_VERSION) && (Java::Java rescue nil)
         count_via_java_runtime
       else
         case RbConfig::CONFIG['host_os']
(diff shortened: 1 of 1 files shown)


lib/rake/cpu_counter.rb
```diff
@@ -32,7 +32,7 @@ def count
     require 'rbconfig'
 
     def count
-      if defined?(Java::Java)
+      if defined?(JRUBY_VERSION) && (Java::Java rescue nil)
```
hsbt: How about to use `RUBY_PLATFORM == 'java'` instead of current imprementation?

jeremyevans: Sure, let me send another diff.

colby-swandale:
@hsbt is this waiting on anything more to merge?

PR #267, opened 2018-06-06
update simple example in README

colby-swandale:
I have an issue with the current simple example on the README being that you can't actually run it unless you create a file located at `test/unittest.rb` which runs a set of tests. This is not good for anyone that is new to Ruby and trying to learn about Rake.

This PR swaps the example with a small task that simply prints to the console. This way, the example Rakefile can be copied and pasted by a user and executed with the expected output. I've also added a paragraph that talks about Rake running a test suite using Rake so that this knowledge is not lost.

### DIFF
--- a/README.rdoc
+++ b/README.rdoc
@@ -45,19 +45,22 @@ Download and install rake with the following.
 First, you must write a "Rakefile" file which contains the build rules. Here's
 a simple example:
 
-  task default: %w[test]
+  task default: %w[hello]
 
-  task :test do
-    ruby "test/unittest.rb"
+  task :hello do
+    puts "Hello World"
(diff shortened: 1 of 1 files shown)


yuki24:
I'm in favor of making the example simpler, but I'd prefer to keep the old one as opposed to replacing it. Rake's Rakefile is not simple and it feels like a big jump from the simplest example to a much more complicated setup.

colby-swandale:
Thanks for the feedback. How would we feel about having an advanced usage guide that explains the complex use cases of Rake instead of having everyone all in the README? It would offer a much better place to explain these use cases and allow us to break them down.

PR #271, opened 2018-08-20
Fixed bug: Reenabled task raises previous exception on second invokation

thorsteneckel:
After a task is `reenable`d an exception of the pre-`reenable` invocation should not be raised but is.

### DIFF
--- a/lib/rake/task.rb
+++ b/lib/rake/task.rb
@@ -141,7 +141,8 @@ module Rake
     # Reenable the task, allowing its tasks to be executed if the task
     # is invoked again.
     def reenable
-      @already_invoked = false
+      @already_invoked      = false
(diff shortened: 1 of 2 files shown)


lib/rake/task.rb
```diff
@@ -141,7 +141,8 @@ def arg_names
     # Reenable the task, allowing its tasks to be executed if the task
     # is invoked again.
     def reenable
-      @already_invoked = false
+      @already_invoked      = false
```
yuki24: I’m not super opinionated about this change, but this would slightly mess up `git blame` and sometimes it may be more annoying than just looking nice.

thorsteneckel:
Thanks for merging @hsbt ❤️

PR #284, opened 2018-11-02
Add rubocop into Travis-CI

colby-swandale:
Note: This PR depends on #282 being merged first.

This PR adds a step in Travis to run rubocop. This would be very beneficial to the project because it makes sure that any new changes to rake conform to the existing style guide.

### DIFF
--- a/.rubocop.yml
+++ b/.rubocop.yml
@@ -4,6 +4,7 @@ AllCops:
   Exclude:
     - doc/**/*.rb
     - rake.gemspec
+    - bin/*
 
 Metrics/LineLength:
   Enabled: true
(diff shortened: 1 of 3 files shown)


colby-swandale:
This has been rebased with `master`

yuki24:
I'd personally prefer to add lint checks to git hooks as it catches syntactical mistakes early. Lint checks on CI create a long feedback loop depending on how many jobs are in the queue.

colby-swandale:
Thanks for the feedback @yuki24. I think both are good practices to have in a project. Having Rubocop in a git hook and having a quick feedback loop is super handy for developers. But also having a final check before any code is merged helps prevent any mistakes from getting into master.

fatkodima:
> Lint checks on CI create a long feedback loop

For this project, `rubocop` should run under several seconds.

fatkodima:
I am trying to run `rubocop` locally and it fails to start. So I'm guessing no one runs it under git hooks also? 😄

PR #286, opened 2018-11-02
run coveralls in CI only

colby-swandale:
When running Rake's test suite, i noticed that the `coveralls` gem is being executed. This doesn't need to be loaded in dev and can be restricted to only load in Travis CI/AppVeyor.


colby-swandale:
This is super-seeded by #289

PR #289, opened 2018-11-05
run coveralls only when COVERALLS env var is present

colby-swandale:
This is to prevent multiple travis-ci jobs running coveralls and results in the coveralls posting multiple reports per PR.

See:
- https://github.com/ruby/rake/pull/288
- https://github.com/ruby/rake/pull/283


colby-swandale:
@hsbt is there anything more i can do to get this merged?

PR #298, opened 2018-12-07
Use Regexp#=== to match loosely

nobu:
Till ruby 2.5, `Object#=~` silently ignored non-string argument, but it has been deprecated since ruby 2.6 and will be removed in the future.  Use `Regexp#===` instead, to match loosely.

Also `/^-/` is not a proper pattern to tell if the argument starts with a `-`.


hsbt:
Fixed with https://github.com/ruby/rake/pull/297

PR #300, opened 2018-12-29
Add ruby 2.6.0 to .travis.yml

colby-swandale:
This PR adds Ruby 2.6 to Travis-CI.

Note that We have to manually update RubyGems because Travis forces each ruby
version to have an old version of RubyGems, which breaks our tests in 2.6.


colby-swandale:
oooohhh, it's because the `.travis.yml` cannot be parsed.

PR #315, opened 2019-05-14
Add tasks.rb to accepted default rake files

svoop:
This is just a suggestion, but as it's ridiculously easy to implement, I'll jump directly to a PR:

Since Bundler is migrating from `Gemfile` to `gems.rb`, how about adding `tasks.rb` as an accepted alternative naming for `Rakefile`?

Another alternative to the existing ones would be `rake.rb`, however, me personally, I prefer to have a functionally descriptive name similar to `gems.rb` – at least for the most essential tools like Bundler and Rake. For less commonly used gems, naming the config file identical to the gem itself makes more sense (e.g. `guard.rb` for Guard).

Thanks for considering this addition!

### DIFF
--- a/lib/rake/application.rb
+++ b/lib/rake/application.rb
@@ -42,7 +42,8 @@ module Rake
       "rakefile",
       "Rakefile",
       "rakefile.rb",
-      "Rakefile.rb"
+      "Rakefile.rb",
+      "tasks.rb"
     ].freeze
 
     # Initialize a Rake::Application object.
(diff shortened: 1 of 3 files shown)


colby-swandale:
Rake already has support for `rakefile.rb`, so i don't really see a clear picture on  why we need this as well.

Also P.S, Bundler is not migrating from `Gemfile` to `gems.rb`, it was just an idea that was being considered.

svoop:
@colby-swandale Ah, you're right, Bundler deprecated Deprecation for the time being :smile: And `rakefile.rb` is totally sufficient, thanks for the hint! Closing.

PR #329, opened 2019-10-04
Skip a taint test on Ruby 2.7

jeremyevans:
Ruby 2.7 is deprecating taint, and taint will no longer have an
effect. See https://bugs.ruby-lang.org/issues/16131.

This just skips the test for FileList#clone and #dup copying the
taint flag on Ruby 2.7+.


hsbt:
👍

PR #331, opened 2019-11-21
Test Rake on TruffleRuby

eregon:
* See https://github.com/oracle/truffleruby/issues/1739
  and f19222ffae9168d4c4d2867f14de06df89febad2.

cc @hsbt

### DIFF
--- a/.github/workflows/macos.yml
+++ b/.github/workflows/macos.yml
@@ -1,6 +1,6 @@
 name: macos
 
-on: [push]
+on: [push, pull_request]
 
 jobs:
   build:
--- a/.github/workflows/ubuntu-rvm.yml
+++ b/.github/workflows/ubuntu-rvm.yml
@@ -7,7 +7,7 @@ jobs:
     runs-on: ubuntu-latest
     strategy:
       matrix:
-        ruby: [ 'jruby-head', 'jruby-9.2.9.0', 'ruby-head', '2.3.8', '2.2.10' ]
+        ruby: [ 'jruby-head', 'jruby-9.2.9.0', 'truffleruby', 'ruby-head', '2.3.8', '2.2.10' ]
     steps:
     - uses: actions/checkout@master
     - name: Set up RVM
--- a/.github/workflows/ubuntu-rvm.yml
+++ b/.github/workflows/ubuntu-rvm.yml
@@ -1,6 +1,6 @@
 name: ubuntu-rvm
 
-on: [push]
+on: [push, pull_request]
 
 jobs:
   build:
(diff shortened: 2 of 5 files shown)


eregon:
One error on this run: https://github.com/ruby/rake/pull/331/checks?check_run_id=314340711
```
  1) Error:
TestRakeClean#test_clean:
NoMethodError: undefined method `desc' for #<#<Class:0x5600>:0x33168>
    /home/runner/work/rake/rake/lib/rake/clean.rb:68:in `<top (required)>'
    /home/runner/work/rake/rake/test/test_rake_clean.rb:7:in `load'
    /home/runner/work/rake/rake/test/test_rake_clean.rb:7:in `block in load'
    /home/runner/work/rake/rake/test/test_rake_clean.rb:7:in `each'
    /home/runner/work/rake/rake/test/test_rake_clean.rb:7:in `load'
    /home/runner/work/rake/rake/test/test_rake_clean.rb:7:in `test_clean'

590 runs, 1629 assertions, 0 failures, 1 errors, 0 skips
```

eregon:
The issue seems to be that `load "rake/clean.rb", true` sets the top-level `self` to `main` in MRI, but to some other class instance on TruffleRuby:
https://github.com/oracle/truffleruby/blob/c3845389c4f38e51283b48a1ea3ccdac4962adbf/src/main/java/org/truffleruby/core/kernel/TruffleKernelNodes.java#L108
cc @chrisseaton

eregon:
That's a trivial fix (use the main object in both cases), and then I can run all `rake` locally successfully.

We'll need a nightly build of TruffleRuby to test in CI though, or skip that `load "rake/clean.rb", true` test until then.

eregon:
It's actually not the same `main` object but some sort of copy of the normal `main`, including modules in the singleton class of the original main.

hsbt:
@eregon Thanks for your work. I'm okay to merge when CI was fixed.

eregon:
@hsbt The CI now passes, with that test slightly modified when running on TruffleRuby 19.3.0, is that acceptable? (18508c4)
I think it would be useful to test TruffleRuby in Rake's CI from now, and not need to wait for the next release.

eregon:
I just merged https://github.com/oracle/truffleruby/commit/f054b32b78ae48d4a24a5e7b164c140fadc36d37, which fixes `load(path, wrap=true)` on TruffleRuby.
But that is not part of the 19.3.0 release of course, it will only be in the 20.0.0 release.

hsbt:
> @hsbt The CI now passes, with that test slightly modified when running on TruffleRuby 19.3.0, is that acceptable? ([18508c4](https://github.com/ruby/rake/commit/18508c4845b1c7026e645295f4ceb27a46308347))

👍

PR #335, opened 2020-02-06
Use ruby/setup-ruby to simplify CI

eregon:
* [ruby/setup-ruby](https://github.com/ruby/setup-ruby) supports more versions than actions/setup-ruby.
* [ruby/setup-ruby](https://github.com/ruby/setup-ruby) is faster to install Rubies than RVM (and much easier in GitHub Actions).

cc @hsbt @ioquatix


eregon:
This currently drops `jruby-head`, do we want to keep it? I can try to support it in https://github.com/ruby/setup-ruby.

eregon:
ruby/setup-ruby also allows testing basically any version on any platform.
Should we test `2.3, 2.2, jruby, truffleruby, ruby-head` on macOS and Windows too?

hsbt:
Thanks! I definitely prefer this.

ioquatix:
Nice work!

PR #339, opened 2020-02-18
New dsl method: lazy(&block)

lionelperrin:
Make possible to lazy define tasks in a namespace. The tasks are
loaded only on lookup.

Lazy definitions can be executed on purpose when using the option
--load-lazy-definitions so that `rake --tasks --load-lazy-definitions`
list all tasks including the one defined in a lazy way.

Example of lazy definitions:
```
namespace 'a' do
  lazy { task 't' }
end
```
or
```
namespace 'a' do
  lazy { load 'namespace_a.rake' }
end
```

### DIFF
--- a/lib/rake/application.rb
+++ b/lib/rake/application.rb
@@ -108,6 +108,7 @@ module Rake
     # Run the top level tasks of a Rake application.
     def top_level
       run_with_threads do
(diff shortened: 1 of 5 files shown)


TSMMark:
This is awesome! I am in need of something like this, and would love to see this merged and released if possible.

PR #340, opened 2020-03-04
v10 use File.open  explicitly backport for CVE-2020-8130

pachkovsky:
Ruby advisory DB was updated (https://github.com/rubysec/ruby-advisory-db/commit/2dd2668e0155c33bb802088b93bdf5b25ae6e9e0) 
It causes bundle-audit to fail for older projects which are unable to update to newer versions of rake.

### DIFF
--- a/History.rdoc
+++ b/History.rdoc
@@ -1,3 +1,9 @@
+=== 10.5.1 / 2020-03-04
+
+Enhancements:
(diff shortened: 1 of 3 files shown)


hsbt:
Unfortunately, v10 is already EOL included security issues.

PR #357, opened 2020-06-12
Simplify default rake test loader

deivid-rodriguez:
Unless I'm missing something, we can require `rake_test_loader` directly. It's also safer, because there's no chance of requiring the file of a different copy of `rake`, and faster.


deivid-rodriguez:
Does any maintainer have any opinion about this?

hsbt:
✋ I'm not sure why the current implementation required. But your changes seems fine.

PR #365, opened 2020-10-17
Limit calls to Dir.chdir in Rake::Application

jonathanhefner:
Ruby [Bug #15661](https://bugs.ruby-lang.org/issues/15661) changed the blockless form of `Dir.chdir` to raise an error if it is called during the block of another `Dir.chdir` call.  This commit refactors code in `Rake::Application` to limit calls to `Dir.chdir`, especially blockless calls.  The eventual goal is to make it safe to call `Rake::Application#run` from within a `Dir.chdir` block.

### DIFF
--- a/lib/rake/application.rb
+++ b/lib/rake/application.rb
@@ -113,7 +113,10 @@ module Rake
         elsif options.show_prereqs
           display_prerequisites
         else
-          top_level_tasks.each { |task_name| invoke_task(task_name) }
+          Dir.chdir(@rakefile_dir || original_dir) do |dir|
+            top_level_tasks.each { |task_name| invoke_task(task_name) }
+            Dir.chdir(dir)
+          end
         end
       end
     end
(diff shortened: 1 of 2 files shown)


lib/rake/application.rb
```diff
@@ -113,7 +113,10 @@ def top_level
         elsif options.show_prereqs
           display_prerequisites
         else
-          top_level_tasks.each { |task_name| invoke_task(task_name) }
+          Dir.chdir(@rakefile_dir || original_dir) do |dir|
+            top_level_tasks.each { |task_name| invoke_task(task_name) }
+            Dir.chdir(dir)
+          end
```
jonathanhefner, line 119: This is based on [my suggestion in Bug #15661](https://bugs.ruby-lang.org/issues/15661#note-15).

lib/rake/application.rb
```diff
@@ -269,18 +272,11 @@ def has_chain?(exception) # :nodoc:
...
-          others = FileList.glob(fn, File::FNM_CASEFOLD)
-          return others.size == 1 ? others.first : fn
-        elsif fn == ""
-          return fn
-        end
+    # Returns first filename from @rakefiles that exists in the specified dir.
+    def have_rakefile(dir = Dir.pwd) # :nodoc:
+      Dir.chdir(dir) do
+        Dir.glob(@rakefiles.map { |name| escape_glob(name) }).first || @rakefiles.find(&:empty?)
       end
```
jonathanhefner, line 279: When support for Ruby < 2.5 is dropped, this `Dir.chdir` can be replaced with [`Dir.glob`'s `base` kwarg](https://docs.ruby-lang.org/en/2.5.0/Dir.html#method-c-glob).

lib/rake/application.rb
```diff
@@ -269,18 +272,11 @@ def has_chain?(exception) # :nodoc:
...
-        if File.exist?(fn)
-          others = FileList.glob(fn, File::FNM_CASEFOLD)
-          return others.size == 1 ? others.first : fn
-        elsif fn == ""
-          return fn
-        end
+    # Returns first filename from @rakefiles that exists in the specified dir.
+    def have_rakefile(dir = Dir.pwd) # :nodoc:
+      Dir.chdir(dir) do
+        Dir.glob(@rakefiles.map { |name| escape_glob(name) }).first || @rakefiles.find(&:empty?)
```
jonathanhefner, line 278: When support for Ruby < 2.5 is dropped, this `Dir.chdir` can be replaced with [`Dir.glob`'s `base` kwarg](https://docs.ruby-lang.org/en/2.5.0/Dir.html#method-c-glob).

PR #368, opened 2020-12-20
Fix breaking change of execution order on TestTask

ysakasin:
Due to #357, execution order on Rake 13.0.2 changes from Rake 13.0.1.

Example:
```ruby
Rake::TestTask do |t|
  t.test_files = [
    "test/a.rb",
    "test/b.rb",
  ]
end
```

On 13.0.2, Rake executes test/b.rb before test/a.rb because test/a.rb are loaded before rake_test_loader.rb.
rake_test_loader.rb executes the Ruby code in ARGV using Kernel.#require, but does not execute test/a.rb which is already loaded.

In addition, Rake 13.0.1 allows specifying test_files without .rb but 13.0.2 doesn't allows.
This commit also fixes this problem.
```ruby
Rake::TestTask do |t|
  t.test_files = [
    "test/setup",
    "test/a.rb",
  ]
end
```

### DIFF
--- a/lib/rake/testtask.rb
+++ b/lib/rake/testtask.rb
@@ -181,7 +181,7 @@ module Rake
       when :testrb
         "-S testrb"
       when :rake
-        "-r#{__dir__}/rake_test_loader"
+        "#{__dir__}/rake_test_loader.rb"
       end
     end
 
--- a/test/test_rake_test_task.rb
+++ b/test/test_rake_test_task.rb
@@ -128,7 +128,7 @@ class TestRakeTestTask < Rake::TestCase # :nodoc:
       t.loader = :rake
     end
 
-    assert_match(/\A-r.*?\Z/, test_task.run_code)
+    assert_includes test_task.run_code, "lib/rake/rake_test_loader.rb"
   ensure
     Gem.loaded_specs["rake"] = rake
   end


mattbrictson:
Thanks for finding this! `rake test` for all of my projects has started failing with rake 13.0.2. I tried the branch referenced in this PR and it fixes the problem. ❤️ 

I hope this gets merged soon.

hsbt:
@ysakasin Thanks to your report and patch. I merged it and release rake-13.0.3 now. Can you confirm to fix your issue with rake-13.0.3?

@mattbrictson Also confirm it with rake-13.0.3?

mattbrictson:
@hsbt confirmed, it is fixed in rake-13.0.3. Thanks!

ysakasin:
@hsbt I confirmed to fix my issue with rake-13.0.3.  Thank you merging and releasing!

evgeni:
Thanks! 13.0.3 also fixes the test issues we've been seeing in our environment.

Would it be a good idea to yank 13.0.2 from Rubygems so that users don't accidentally install it?

deivid-rodriguez:
Apologies for the unintentional breaking change and thanks @ysakasin for the quick fix, @hsbt for the quick release and @mattbrictson for the quick testing :heart:.

@evgeni No, that's not a good idea. The use case for yanking a gem is if someone stole your credentials and pushed a malicious version of your gem, or if you accidentally pushed sensitive data together with your gem, or things like that. Yanking a gem has bad consequences on lockfiles locked to the yanked version and in general is not recommended for deleting gems that have bugs. Imagine if every gem that has a bug needed to be yanked, we'd be wiping out the whole rubygems.org DB :rofl:.

PR #369, opened 2020-12-21
Fix 13.0.2 changelog entry

ShockwaveNN:
According to [this](https://github.com/ruby/rake/commit/6b8c70d2b39ac7c952f446d82fcf5e2fe6a09e09) commint name those changes are for 13.0.2

Seems this is a typo, since current history.rdoc contains two entries for 13.0.1 and no for 13.0.2


hsbt:
Thanks, I was fixed at https://github.com/ruby/rake/commit/c2eeae2fe2b67170472a1441ebf84d3a238c3361

PR #370, opened 2020-12-21
Add ability to use named arguments when invoking a task

### LINKED ISSUE #81: Support named task arguments
When creating tasks that take multiple optional arguments, specifying parameters becomes tedious.

When calling the following task

    task :test, [:database, :table, :column, :host, ;port, :user, :password] do
        # Something
    end

and specifying only parameter `user`, I have to count its position and call rake as follows:

    rake 'test[,,,,,testuser,]'

I suggest supporting a syntax using named parameters that makes calling these tasks easier and more readable, e.g.

    rake 'test{user:testuser}'

I'd like to put this up for discussion, if it is acceptable, I'm willing to create a pull request.

fatkodima:
Closes #81

Syntax:
```
rake 'test{user:testuser, password:secret}'
```

### DIFF
--- a/lib/rake/application.rb
+++ b/lib/rake/application.rb
@@ -161,25 +161,62 @@ module Rake
     end
 
     def parse_task_string(string) # :nodoc:
-      /^([^\[]+)(?:\[(.*)\])$/ =~ string.to_s
-
-      name           = $1
-      remaining_args = $2
-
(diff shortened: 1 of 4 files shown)


fatkodima:
@hsbt @nobu It's been a while since this PR was created. Can you, please, review?

PR #372, opened 2020-12-23
Preserve match_data in rules

### LINKED ISSUE #141: Idea: Preserve matchdata in rules
Here's part of a rule I just wrote the other day:

```ruby
stampfile_pattern = %r"^members/(?<email>.*)/\d{2}-(?<step>\w+).timestamp$"
rule stampfile_pattern do |t|
  match           = stampfile_pattern.match(t.name)
  email           = match[:email]
  step            = match[:step]
  # ...
end
```

Notice how, after it matches, I have to re-apply the regex inside the rule definition in order to pull out named matchdata.

I was thinking of introducing something like this as an alternative:

```ruby
stampfile_pattern = %r"^members/(?<email>.*)/\d{2}-(?<step>\w+).timestamp$"
rule stampfile_pattern do |t|
  email           = t.matchdata[:email]
  step            = t.matchdata[:step]
  # ...
end
```

Before I dive in, what do you think? And is there another way to accomplish this that I've missed?

fatkodima:
Closes #141 

Example of usage (from the issue):
```ruby
rule /^members/(?<email>.*)/\d{2}-(?<step>\w+).timestamp$/ do |t|
  email = t.match_data[:email]
  step  = t.match_data[:step]
  # ...
end
```

cc/ @avdi

### DIFF
--- a/lib/rake/task.rb
+++ b/lib/rake/task.rb
@@ -38,6 +38,10 @@ module Rake
     # will be skipped unless you reenable them.
     attr_reader :already_invoked
 
+    # MatchData object returned when matching the task name
+    # against the rule pattern.
+    attr_accessor :match_data
+
     # Return task name
     def to_s
       name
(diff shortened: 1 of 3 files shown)


avdi:
Cool!!!

PR #374, opened 2021-02-03
Lazily load `set`

deivid-rodriguez:
This file seems required by `rake` by default. By lazily loading `set`, usages of `rake` that don't use this part of `rake`, don't unnecessarily activate the `set` gem.

This seems really minor, I know. But when testing rubygems & bundler, it's interesting for us to be in full control of the gems that are
loaded, to make sure that we don't unintentionally activate gems causing our users to be unable to select the version of those gems that they need.

### DIFF
--- a/lib/rake/thread_pool.rb
+++ b/lib/rake/thread_pool.rb
@@ -1,5 +1,4 @@
 # frozen_string_literal: true
-require "set"
 
 require "rake/promise"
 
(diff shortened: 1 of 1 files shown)


lib/rake/thread_pool.rb
```diff
@@ -10,6 +9,7 @@ class ThreadPool # :nodoc: all
     # Creates a ThreadPool object.  The +thread_count+ parameter is the size
     # of the pool.
     def initialize(thread_count)
+      require "set"
```
casperisfine, line 12: Honestly, looking at the usage the use of a set doesn't seem particularly useful. An Array would perform just I good I think.

The only operation that might be slower is `@threads.delete Thread.current`, but unless you have a huge amount of thread you probably won't notice any difference.

deivid-rodriguez: I'm happy to change this to plain arrays too.

PR #375, opened 2021-02-09
Add safe invocation CLI syntax

casperisfine:
Example: `rake may:or:maynot:exist?`

Such syntax is useful for projects like buildpacks and other automated CI/CD tools.

It would allow these tools to very easily invoke a taskif it exists, without failing if it doesn't.

@hsbt is this a feature that could be acceptable?

Also cc @schneems, since as maintainer of https://github.com/heroku/heroku-buildpack-ruby I suspect such feature could be useful.

### DIFF
--- a/lib/rake/application.rb
+++ b/lib/rake/application.rb
@@ -155,19 +155,24 @@ module Rake
 
     # Invokes a task with arguments that are extracted from +task_string+
     def invoke_task(task_string) # :nodoc:
-      name, args = parse_task_string(task_string)
+      name, args, safe_call = parse_task_string(task_string)
+      return if safe_call && !lookup(name)
+
       t = self[name]
       t.invoke(*args)
     end
 
     def parse_task_string(string) # :nodoc:
-      /^([^\[]+)(?:\[(.*)\])$/ =~ string.to_s
+      string = string.to_s.dup
+      safe_call = !!string.chomp!('?')
+
+      /^([^\[]+)(?:\[(.*)\])$/ =~ string
 
       name           = $1
       remaining_args = $2
 
-      return string, [] unless name
-      return name,   [] if     remaining_args.empty?
+      return string, [], safe_call unless name
+      return name,   [], safe_call if remaining_args.empty?
 
       args = []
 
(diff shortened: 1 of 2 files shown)


lib/rake/application.rb
```diff
@@ -155,19 +155,24 @@ def thread_pool             # :nodoc:
...
 
     def parse_task_string(string) # :nodoc:
-      /^([^\[]+)(?:\[(.*)\])$/ =~ string.to_s
+      string = string.to_s.dup
+      safe_call = !!string.chomp!('?')
+
+      /^([^\[]+)(?:\[(.*)\])$/ =~ string
 
       name           = $1
       remaining_args = $2
```
nobu, line 172:
```suggestion
      /^([^\[]+)(?:\[(.*)\])(\?)?$/ =~ string.to_s

      name           = $1
      remaining_args = $2
      safe_call      = !!$~.begin(3)
```

casperisfine: @nobu I tried that but it doesn't work for various reasons. Mostly because simple tasks without arguments actually don't match that pattern.

hsbt:
@byroot With this change, a task whose name ends with `?` and an argument ending with `?` (e.g. `rake "search[what?]"`) silently stop working, since the trailing `?` is chomped before parsing.

I have rarely seen such names in real projects, so I could accept this as an incompatibility, but is that breakage intended, or do you plan to handle these cases?

byroot:
> but is that breakage intended,

I didn't think those were valid names. If it is a concern, I can look at another symbol that isn't currently a valid name if there is one.

But as you suggested, I can also look at handling the ambiguity of it. Just let me know what you prefer.

hsbt:
Sorry for the slow reply. Let's keep `?` and handle the ambiguity. Defining a list of valid task name characters now would mean adding a fallback every time someone turns out to be using one, so I'd rather cover just these two cases and leave the rest alone.

byroot:
@hsbt no worries. Reopened at https://github.com/ruby/rake/pull/765 with the requested changed (because this PR was done with a deprecated account).

Issue #382, opened 2021-04-07
Update the gh-pages branch


tamouse:
Out of date links to Avdi's rake tutorial

hsbt:
I've generated that from https://github.com/ruby/rake/commit/a07e637c080d8674cd2e1da26c51aaacb67b2d80

PR #383, opened 2021-04-07
Update documentation

tamouse:
I noticed that the index.html at https://ruby.github.io/rake/ contained out of date links to @avdi 's tutorials. I thought this would be a pretty simple update on gh-pages, but it looks like a lot of files were out of date. 🙍‍♀️


hsbt:
https://github.com/ruby/rake/issues/382#issuecomment-874503956

Issue #173, opened 2016-11-10
How to get exit status of the rake task using rake API


saraniyadev:
Is their a possibility to get the exit status of a rake task when executed inside a ruby script. 

My config rake method looks like below 
`desc "Run serverspec to #{host}"
    RSpec::Core::RakeTask.new(host) do |t|
      ENV['TARGET_HOST'] = host
      t.pattern = "spec/cfengine3/*_spec.rb"
    end`

My run methods contains below code 
Rake.application.init
Rake.application.top_level

I am not able to get exit status of my rake task invoked by top_level

Issue #192, opened 2017-02-10
find the Status of each rake tasks


anoobbava:
Hi,

I had too many rake tasks in the project. Need to find whether the rake execution is success or not.  I have used rakefile, it will helps for the rake file name, args, start and end time. But whether the execution is sucess  or any exception occurs we can  not find out.Is there anyway find out. Posted similar question in Stackoverflow.

http://stackoverflow.com/questions/42140463/find-the-exception-from-rakefile-in-rake-tasks

hsbt:
Is it similar with #173 ?

I understood your request and usecase ( also #173 ). But I could not better API name like "RakeTask#status" or "RakeTask#exit_status"

anoobbava:
@hsbt. Here also we can't find the status of last take success or not.  Is it needed right ?

karreiro:
Hey everyone, I've opened this PR https://github.com/ruby/rake/pull/420 targeting to solve this issue.

I'd love to know what you think about it :) I'd be happy to change it to a different approach if you consider a good idea.

anoobbava:
@karreiro Thanks a Ton man 🙏🏽 , will wait for this change to be merged 👯

PR #420, opened 2022-05-08
Fix the issues #173 and #192 with the 'Rake::Task#results'

karreiro:
This PR fixes issues #173 and #192 by introducing the `Rake::Task#results` method.

We could also have the `RakeTask#status` or `RakeTask#exit_status` methods (and maybe keep `#results` private).

However, we would need to inspect the result of each task execution and adopt a convention for statuses. Thus, exposing `#results` seems a bit more flexible and permits this kind of use case:

```ruby
task default: %w[test]

task :test do
  puts 'Testing...'
  :success
end
```

```ruby
require "rake"

Rake.application.load_rakefile

task = Rake::Task[:test]

task.invoke

p task.results
# => [:success]
```


karreiro:
(I just rebased the branch) 👀

PR #423, opened 2022-05-09
Apply RuboCop linting for Ruby 2.3

tnir:
- Pin rubocop to 0.81.0 to keep [TargetRubyVersion: 2.3](https://github.com/ruby/rake/blob/881416e322d597e47e62935ef6a9dbaed934fbfc/.rubocop.yml#L2)
- Update `.rubocop.yml`:
  - Rename `Metrics/LineLength` to `Layout/LineLength` ([0.78.0](https://github.com/rubocop/rubocop/blob/master/CHANGELOG.md#0780-2019-12-18))
  - Remove `Style/BracesAroundHashParameters` cop` ([0.80.0](https://github.com/rubocop/rubocop/blob/master/CHANGELOG.md#0800-2020-02-18))
  - Rename `Layout/TrailingBlankLines` to `Layout/TrailingEmptyLines` (older version...)

Signed-off-by: Takuya Noguchi <takninnovationresearch@gmail.com>

PR #426, opened 2022-05-10
Add RuboCop job to Actions

tnir:
Just to avoid drifts such as ones fixed in #423.

Signed-off-by: Takuya Noguchi <takninnovationresearch@gmail.com>

### DIFF
--- a/.github/workflows/lint.yml
+++ b/.github/workflows/lint.yml
@@ -1,0 +1,16 @@
+name: lint
+
+on: [push, pull_request]
+
+jobs:
+  lint:
+    runs-on: ubuntu-latest
+    continue-on-error: true
+    steps:
+    - uses: actions/checkout@v3
+    - uses: ruby/setup-ruby@v1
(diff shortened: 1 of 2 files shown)


tnir:
Looks fine to me as far as I see https://github.com/ruby/rake/runs/6703247369?check_suite_focus=true. @hsbt What do you think?

hsbt:
I don't like to add lint to CI mandatory.

It's good to auto-fix each commits in the pull-requests. But I'm not sure how to enable it on the GHA.

tnir:
> I don't like to add lint to CI mandatory.

Just added `continue-on-error: true` to the job not to mandate it.

> It's good to auto-fix each commits in the pull-requests. But I'm not sure how to enable it on the GHA.

Sounds good but RuboCop auto corrections for all cops are not always available.

hsbt (approved):

hsbt:
Thanks!

PR #440, opened 2022-07-18
Eagerly require set in thread_pool.rb

jeremyevans:
Lazily requiring it in this matter can break rake whenever
a rake task limits access to the file system.  For example:

```ruby
task :default do
  Dir.chroot Dir.pwd
end
```

One reason to lazily require this is to save on memory, but
since Rake::Application appears to always use a thread pool
(Rake::Application#run -> top_level -> run_with_threads ->
thread_pool), it doesn't looks like lazily requiring actually
saves memory in this case.

### DIFF
--- a/lib/rake/thread_pool.rb
+++ b/lib/rake/thread_pool.rb
@@ -1,6 +1,7 @@
 # frozen_string_literal: true
 
 require "rake/promise"
+require "set"
 
 module Rake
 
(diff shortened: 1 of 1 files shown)


hsbt:
@jeremyevans Thanks. CI is broken with https://github.com/ruby/rake/runs/7395726695?check_suite_focus=true#step:4:6. I fix it and rebase this PR.

PR #444, opened 2022-08-15
Correct RuboCop offenses

deivid-rodriguez:
Introduced after #147, which was not fully rebased when it was merged.


hsbt:
Nice catch!

PR #468, opened 2023-02-01
Generate gh-pages for most recent Rake

zzak:
Using RDoc 2.5.0 from ruby-trunk


hsbt:
We leaved from `gh-pages` deployment at https://github.com/ruby/rake/pull/477

PR #483, opened 2023-03-21
silence warnings during execution of rake tasks in Rakefile (ex: rake test)

luke-gru:


hsbt:
I'm not sure what you want to ignore warnings. Can you share warnings that are suppressed?

PR #515, opened 2023-09-18
Pattern matching support for arguments

rgarner:
Implement `Rake::TaskArguments#deconstruct_keys`. This means in an idiomatic ruby 3.x rake task we can use rightward assignment to say:

```
task :get, %i[tenant id] do |_t, args|
  args => {tenant:, id:}
  ...
end
```

... and omit the `.to_h` from `args`, raising `NoMatchingPatternError` if either of the two params is absent from the task args.

### DIFF
--- a/lib/rake/task_arguments.rb
+++ b/lib/rake/task_arguments.rb
@@ -94,6 +94,10 @@ module Rake
       @hash.fetch(*args, &block)
     end
 
+    def deconstruct_keys(keys)
+      @hash.slice(*keys)
+    end
+
     protected
 
     def lookup(name) # :nodoc:
--- a/test/test_rake_task_arguments.rb
+++ b/test/test_rake_task_arguments.rb
@@ -54,6 +54,13 @@ class TestRakeTaskArguments < Rake::TestCase # :nodoc:
     assert_equal 0,  h.fetch(:one)
   end
 
+  def test_deconstruct_keys
+    omit "No stable pattern matching until Ruby 3.1 (testing #{RUBY_VERSION})" if RUBY_VERSION < "3.1"
+
+    ta = Rake::TaskArguments.new([:a, :b, :c], [1, 2, 3])
+    assert_equal ta.deconstruct_keys([:a, :b]), { a: 1, b: 2 }
+  end
+
   def test_enumerable_behavior
     ta = Rake::TaskArguments.new([:a, :b, :c], [1, 2, 3])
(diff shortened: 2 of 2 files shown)


test/test_rake_task_arguments.rb
```diff
@@ -54,6 +54,11 @@ def test_to_hash
     assert_equal 0,  h.fetch(:one)
   end
 
+  def test_deconstruct_keys
```
nobu, line 57: You should omit older versions.
I'm not sure if ruby 3.0 should be supported, because one-line pattern matching was experimental.

```suggestion
  def test_deconstruct_keys
    omit "Pattern matching is not supported yet" if RUBY_VERSION < "3.1"
```

rgarner: Agreed on both counts with thanks, updated and rebased

nevans (commented):
`#deconstruct_keys` should handle `keys == nil`, or `**rest` will be broken.

```ruby
# normal behavior:
{a: 1, b: 2, c: 3} => {a:, **rest}
a     # => 1
rest  # => {b: 2, c: 3}

# without handling `keys == nil`
class Example
   def initialize(hash) = @hash = hash
   def deconstruct_keys(keys) = @hash.slice(*keys)
end

Example.new({a: 1, b: 2, c: 3}) => {a:, **rest}
# !=> #<Example:0x00007479b6b781b8 @hash={a: 1, b: 2, c: 3}>: key not found: :a (NoMatchingPatternKeyError)
```

_(My apologies for not creating a PR... I can do that later today.)_

nevans:
Okay, I did create a small PR to handle the nil case: #635.

Thanks for this PR, @rgarner!  I'm a big fan of adding `#deconstruct` and/or `#deconstruct_keys` everywhere they make sense.  🙂

rgarner:
Great shout @nevans, thanks for spotting. I rarely use the `**rest` case and was unaware of the `nil` semantics but will take this lesson with me!

lib/rake/task_arguments.rb
```diff
@@ -94,6 +94,10 @@ def fetch(*args, &block)
       @hash.fetch(*args, &block)
     end
 
+    def deconstruct_keys(keys)
+      @hash.slice(*keys)
+    end
```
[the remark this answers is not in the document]
Yltrik: ?

PR #545, opened 2024-03-14
Use Struct instead of OpenStruct.

hsbt:
I will migrate `osturct` as bundled gems at Ruby 3.5. We should remove its dependency from Rake.

### DIFF
--- a/lib/rake/application.rb
+++ b/lib/rake/application.rb
@@ -165,9 +165,7 @@ module Rake
 
     # Application options from the command line
     def options
-      return @options if @options
-
-      @options = Struct.new(
+      @options ||= Struct.new(
         :always_multitask, :backtrace, :build_all, :dryrun,
         :ignore_deprecate, :ignore_system, :job_stats, :load_system,
         :nosearch, :rakelib, :show_all_tasks, :show_prereqs,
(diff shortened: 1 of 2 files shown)


lib/rake/application.rb
```diff
@@ -165,7 +165,13 @@ def add_loader(ext, loader)
...
     # Application options from the command line
     def options
-      @options ||= OpenStruct.new
+      @options ||= Struct.new(
+        :always_multitask, :backtrace, :build_all, :dryrun,
+        :ignore_deprecate, :ignore_system, :job_stats, :load_system,
+        :nosearch, :rakelib, :show_all_tasks, :show_prereqs,
+        :show_task_pattern, :show_tasks, :silent, :suppress_backtrace_pattern,
+        :thread_pool_size, :trace, :trace_output, :trace_rules
+      ).new
```
zenspider, line 174: It would be nicer if this was a named class:

```
NoMethodError: undefined method `verbose' for an instance of #<Class:0x00000001038112e0>
```

vs something like:

```
NoMethodError: undefined method `verbose' for an instance of #<Rake::Options:0x00000001038112e0>
```

Issue #272, opened 2018-08-30
Aborting `rake` execution needlessly dumps a backtrace


booch:
When aborting `rake` with Ctrl+C, it outputs `rake aborted!` then dumps a long backtrace. This seems unnecessary, as aborting isn't really an unexpected outcome of pressing Ctrl+C (or sending an INT signal).

I'd like to propose catching the `Interrupt` exception and exiting cleanly (with an exit code of 1). We can still print the backtrace if the `--trace` option is used.

I can submit a pull request, if this seems reasonable.

ret394:
`ruby 2.7.1p83 (2020-03-31 revision a0c7c23c9c) [x64-mingw32]` & rake also dumps Traceback.

thanks. have a nice day. keep up the good work.

[ret394](https://github.com/ret394)

mateusdeap:
Is this something we still want to look at? I can also submit a PR for this if needed.

SwagDevOps:
Yes! I'm still interested.

PR #561, opened 2024-04-29
Handle for SIGTERM

shayonj:
In containerized workloads, it's common to receive a `SIGTERM` signal when a pod or task is shutting down. Currently, when this occurs, the `rake` task throws a `SignalException` and returns a status code of `1`. This behavior can be misinterpreted by container orchestration systems (like Fargate, Kubernetes, etc.) as an error, rather than a graceful exit.

This PR proposes that, when a `SIGTERM` signal is received by a `rake` task, it should return an exit code of `143`. This exit code is the standard response for a `SIGTERM`, indicating a proper shutdown was acknowledged and handled gracefully.

I only added this for SIGTERM, but should be extensive for other handlers too if needed in future. 

Related: https://github.com/ruby/rake/issues/272

### DIFF
--- a/lib/rake/application.rb
+++ b/lib/rake/application.rb
@@ -94,6 +94,8 @@ module Rake
           # Backward compatibility for capistrano
           args = handle_options
         end
+
+        setup_signal_handling
         load_debug_at_stop_feature
         collect_command_line_tasks(args)
       end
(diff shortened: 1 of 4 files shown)


shayonj:
👋🏾 @hsbt thanks for taking a look, just updated the branch. Let me know if you have any feedback 🙏🏾

PR #562, opened 2024-04-30
Use macos-13 for old ruby versions

hsbt:


hsbt:
I try with https://github.com/ruby/rake/pull/563

PR #566, opened 2024-05-09
Use `require_relative` in the Rake codebase

koic:
If there are many searches in the `$LOAD_PATH` in the user environment, require will perform unnecessary searches that are not needed. In contrast, `require_relative` is efficient because it uses a relative path.

Rake requires Ruby 2.3+, it is possible to use `require_relative`, which was introduced in Ruby 1.9.
https://github.com/ruby/rake/blob/v13.2.1/rake.gemspec#L99


hsbt:
Thanks!

PR #567, opened 2024-05-11
Fix Rake.verbose to return true/false even if unset

zenspider:
Unset returns FileUtilsExt.verbose_flag which is set to Object.new by
default, which is truthy. Compare against DEFAULT and return
verbose_flag only if set.

### DIFF
--- a/lib/rake/file_utils_ext.rb
+++ b/lib/rake/file_utils_ext.rb
@@ -60,7 +60,7 @@ module Rake
           FileUtilsExt.verbose_flag = oldvalue
         end
       end
-      FileUtilsExt.verbose_flag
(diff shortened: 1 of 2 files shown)


zenspider:
@hsbt comment pls?

jasonkarns:
Additionally `Rake.verbose?` would be nice?

jasonkarns:
Is there any chance of this getting forward progress? It seems I end up here every few months when I try to use the `verbose` method to interrogate rake's rebose flag.

(current example: factory bot's linter:

```rb
task :lint, [:factories] => :environment do |t, args|
  # [/snip]
  FactoryBot.lint(factories, verbose:) # incorrectly sets linter to verbose because the default flag object is truthy
end
```

PR #573, opened 2024-06-08
Remove dependency on `win32ole`

Earlopain:
This gem is being bundled in Ruby 3.5: https://github.com/ruby/ruby/commit/f365bef0c78085b7db9b6736ae59c1657baa79a3

So, just use buildin commands to do the same thing.

The powershell version will work on at least Windows 8 (from 2012). There are no docs I could find for earlier versions, so it might work on even earlier versions.  https://learn.microsoft.com/en-us/powershell/scripting/install/powershell-support-lifecycle?view=powershell-7.4. Fall back to the deprecated wmic if it doesn't exist.

I have openen https://github.com/grosser/parallel/pull/346 and https://github.com/ruby-concurrency/concurrent-ruby/pull/1051 which do the same thing.

### DIFF
--- a/lib/rake/cpu_counter.rb
+++ b/lib/rake/cpu_counter.rb
@@ -58,11 +58,20 @@ unless Rake::CpuCounter.method_defined?(:count)
     end
 
     def count_via_win32
-      require 'win32ole'
-      wmi = WIN32OLE.connect("winmgmts://")
-      cpu = wmi.ExecQuery("select NumberOfCores from Win32_Processor") # TODO count hyper-threaded in this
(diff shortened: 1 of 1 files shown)


hsbt:
Thanks 👍  I confirmed your powershell script is working correctly.

```
hsbt@FISCHER C:\Users\hsbt\source\ruby.build\2022>powershell -command "Get-CimInstance -ClassName Win32_Processor -Property NumberOfCores | Select-Object -Property NumberOfCores"

NumberOfCores
-------------
            8
```

PR #581, opened 2024-07-08
Interrupt rescuing causes issues with debugger

notEthan:
The way Rake rescues exceptions running a task, I think, rescues too much. The problem I encounter is when I am using [debug](https://github.com/ruby/debug) in a test task that runs in a subprocess (via `Kernal#system`) - raising an Interrupt with ctrl+c leaves the terminal in a state where it no longer echoes input.

To reproduce, with the `debug` gem installed and this Rakefile:

```ruby
task(:default) { system("ruby -r debug -e debugger") }
```

Run `rake`; at the debug prompt hit ctrl+c. Your terminal no longer echoes your input; the tty is left in raw mode (I eventually found you can fix this by running `stty -raw`)

I think the solution to this is what RSpec does, rescuing broadly but with some exceptions to the Exceptions - see https://github.com/rspec/rspec-support/blob/v3.13.0/lib/rspec/support.rb#L145-L153

I've copied that in this PR and replaced every `rescue Exception` in `lib/`, though only lib/rake/application.rb affects my usage, I think it should be correct in the others as well.

### DIFF
--- a/lib/rake.rb
+++ b/lib/rake.rb
@@ -21,7 +21,17 @@
 # IN THE SOFTWARE.
 #++
 
-module Rake; end
+module Rake
+  module AllExceptionsExceptOnesWeMustNotRescue
+    # These exceptions are dangerous to rescue as rescuing them
+    # would interfere with things we should not interfere with.
+    AVOID_RESCUING = [NoMemoryError, SignalException, Interrupt, SystemExit]
+
+    def self.===(exception)
+      AVOID_RESCUING.none? { |ar| ar === exception }
+    end
+  end
+end
 
 require_relative "rake/version"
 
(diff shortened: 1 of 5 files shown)


lib/rake.rb
```diff
@@ -21,7 +21,17 @@
 # IN THE SOFTWARE.
 #++
 
-module Rake; end
+module Rake
+  module AllExceptionsExceptOnesWeMustNotRescue
```
notEthan, line 25: I wasn't sure where would be a good place for this, just put it here for a draft PR.

notEthan:
I keep encountering this, changed draft PR to ready, hoping it looks acceptable.

PR #589, opened 2024-10-01
chore: refactor/reformat the heredocs (in tests) ...

pvdb:
... by consistently using the squiggly heredoc syntax (using `<<~` instead of `<<-`) and by fixing the indentation of all multi-line heredoc content accordingly.

Since 704c2f10667b4d244191e1bff50d92f34a0e861c (April 2023), `rake` has been targeting `ruby v2.3` _(and above)_, which is the `ruby` version that introduced @avdi's suggested [squiggly heredoc syntax](https://avdi.codes/about-the-ruby-squiggly-heredoc-syntax/), so there should be no issue with using it throughout the `rake` codebase.

### DIFF
--- a/test/helper.rb
+++ b/test/helper.rb
@@ -96,10 +96,10 @@ class Rake::TestCase < Test::Unit::TestCase
     FileUtils.mkdir_p @system_dir
 
     open File.join(@system_dir, "sys1.rake"), "w" do |io|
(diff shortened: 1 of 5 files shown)


hsbt:
This change is reasonable and make more readable. Thanks!

PR #590, opened 2024-10-01
chore: refactor how temporary files are created (in tests)

pvdb:
... by consistently using [`File::write`](https://ruby-doc.org/3.3.5/IO.html#method-c-write) instead of [`Kernel#open`](https://ruby-doc.org/3.3.5/Kernel.html#method-i-open).

At the moment, most temporary files (in tests) are written using `Kernel#open` even though sometimes `File::write` is already being used;  this PR refactors all tests to consistently use `File::write`.

It's possibly a matter of style/preference/opinion, but this also makes things slightly more readable and slightly less convoluted; either way, I think `File.write` is a tad more "intention revealing" than `Kernel#open` as well: we are "writing a (Rake-)file" as opposed to "opening a file for writing" ... small but subtle difference. :sweat_smile:

In addition, the content of temporary files is  refactored to consistently use heredocs, instead of a _(quite random)_ mix of heredocs and [`IO#<<`](https://ruby-doc.org/3.3.5/IO.html#method-i-3C-3C) ... double win! :smiley:

PS - so as to not introduce any different behaviour, and to keep things functionally identical, I introduced a couple of newline characters (`"\n"`) even though these appear to be immaterial _(the tests run green with and without them)_ so I'm happy to take them out again.

### DIFF
--- a/test/helper.rb
+++ b/test/helper.rb
@@ -95,21 +95,17 @@ class Rake::TestCase < Test::Unit::TestCase
 
     FileUtils.mkdir_p @system_dir
 
-    open File.join(@system_dir, "sys1.rake"), "w" do |io|
-      io << <<~SYS
-        task "sys1" do
-          puts "SYS1"
-        end
-      SYS
-    end
+    File.write File.join(@system_dir, "sys1.rake"), <<~SYS
+      task "sys1" do
+        puts "SYS1"
+      end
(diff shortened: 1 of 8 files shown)


hemalvarambhia:
:+1:

hsbt:
👍  Thank you.

PR #66, opened 2015-09-07
Historical info on Jim, acknowledgement and thanks

jaustinhughey:
This is just a friendly (albeit somber) pull request to acknowledge Jim Weirich in a more salient, respectful way. The original copyright statement under "other stuff", regardless of who put it there, now seems almost disrespectful, as if he's an afterthought and nothing more.

It's important that we acknowledge and thank innovators for their work, even in death; such ongoing acknowledgement shows veterans and newcomers alike that building great things leaves a lasting legacy; seeing that legacy can inspire others to build *their own* legacy by contributing to the community.

More functionally, newcomers to Ruby might see lots of references to Jim and jimweirich/rake as the repository location (e.g. books, old articles/blog posts, etc.), so this may clear up some confusion.

Merely a friendly suggestion; please discuss or ask questions if you have them. Thanks!

PR #600, opened 2024-10-21
Revert "Link to Jim's last `rake` commit (not the git tree with that SHA)"

pvdb:
Reverts ruby/rake#593 - as per @jahio's https://github.com/ruby/rake/pull/66#issuecomment-2395153316 and clarification, I now better understand his original intent, so I think it makes sense to revert to the original link ... @hsbt - do you agree?


hsbt:
Thank you for your help.

But I prefer the current revision. So, I'll close this.

jahio:
I'm ok with it too. Important thing is remembering Jim.

PR #635, opened 2025-06-02
Fix TaskArguments#deconstruct_keys with keys = nil

nevans:
This fixes a bug in the initial implementation (#515):

`#deconstruct_keys` should handle `keys == nil`, or `**rest` will be broken.

For example, the normal behavior looks like this:
```ruby
  {a: 1, b: 2, c: 3} => {a:, **rest}
  a     # => 1
  rest  # => {b: 2, c: 3}
```

But without handling `keys == nil`, we'll get this:
```ruby
  class Example
    def initialize(hash) = @hash = hash
    def deconstruct_keys(keys) = @hash.slice(*keys)
  end

  Example.new({a: 1, b: 2, c: 3}) => {a:, **rest}
  # !> "#{inspect}: key not found: :a" (NoMatchingPatternKeyError)
```

### DIFF
--- a/test/test_rake_task_arguments.rb
+++ b/test/test_rake_task_arguments.rb
@@ -58,6 +58,7 @@ class TestRakeTaskArguments < Rake::TestCase # :nodoc:
     omit "No stable pattern matching until Ruby 3.1 (testing #{RUBY_VERSION})" if RUBY_VERSION < "3.1"
 
     ta = Rake::TaskArguments.new([:a, :b, :c], [1, 2, 3])
+    assert_equal ta.deconstruct_keys(nil), { a: 1, b: 2, c: 3 }
     assert_equal ta.deconstruct_keys([:a, :b]), { a: 1, b: 2 }
   end
 
--- a/lib/rake/task_arguments.rb
+++ b/lib/rake/task_arguments.rb
@@ -95,7 +95,7 @@ module Rake
     end
 
     def deconstruct_keys(keys)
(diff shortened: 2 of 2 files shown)


test/test_rake_task_arguments.rb
```diff
@@ -58,6 +58,7 @@ def test_deconstruct_keys
     omit "No stable pattern matching until Ruby 3.1 (testing #{RUBY_VERSION})" if RUBY_VERSION < "3.1"
 
     ta = Rake::TaskArguments.new([:a, :b, :c], [1, 2, 3])
+    assert_equal ta.deconstruct_keys(nil), { a: 1, b: 2, c: 3 }
     assert_equal ta.deconstruct_keys([:a, :b]), { a: 1, b: 2 }
```
rgarner, line 62:
```suggestion
    assert_equal({ a: 1, b: 2, c: 3 }, ta.deconstruct_keys(nil))
    assert_equal({ a: 1, b: 2 }, ta.deconstruct_keys([:a, :b]))
```

Being an RSpec devotee, I spoke with an accent when I added this test and had the `assert_equal(expected, actual)` order backwards. It'd only affect the failure message but I'd hate for someone plugging a hole to copy my bad accent...

nevans: I'm also an rspec guy, so I get the order of these backwards pretty often myself.  Anyway, I pushed an update with them in the correct order.  Thanks. 🙂

PR #638, opened 2025-06-06
Add --console 

imanpalsingh:
This PR adds a new` --console` flag that starts an IRB session with the Rake environment loaded.

This is my first contribution to this project. Please let me know if I missed anything. I am not sure if tests are required for this. If they are, I would appreciate guidance on what to cover.


hsbt:
You can do it with `rake -e 'binding.irb'`

PR #713, opened 2026-04-03
Dedupe and simplify `standard_system_dir`

pvdb:
[released](https://released.blabberate.com/p/ruby/rake/713) **Dedupe and simplify `standard_system_dir`**

---

Hot on the heels of #669, this small refactor dedupes the multiple method definitions by inlining the `Win32.windows?` check, which means we don't end up with two different method implementations for the same thing, which is a tad confusing for e.g. `ctags`:

ctags

It also replaces the duplicated `Win32.windows?` check with the existing `Rake::Application.windows?` utility function, already used elsewhere in the class.

This in turn allowed me to add two explicit tests, one for the Windows behaviour, another one for the Unix behaviour; it does this by stubbing the `Rake::Application.windows?` utility function, which is indeed a tad strange, but both tests are modelled after the existing `test_load_from_calculated_system_rakefile` _(which is arguably even stranger, as it needs to stub a private function in order to do its thing)_ :smiley:

### DIFF
--- a/lib/rake/application.rb
+++ b/lib/rake/application.rb
@@ -754,12 +754,10 @@ module Rake
     end
 
     # The standard directory containing system wide rake files.
-    if Win32.windows?
-      def standard_system_dir #:nodoc:
+    def standard_system_dir #:nodoc:
+      if windows?
         File.join(Dir.home, "Rake")
-      end
-    else
-      def standard_system_dir #:nodoc:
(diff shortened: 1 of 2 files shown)


hsbt (approved):

pvdb:
Thanks, @hsbt ! :pray:

PR #729, opened 2026-04-20
refactor: move platform detection to `Rake` module, and delete `Rake::Win32` module

pvdb:
The `Rake::Win32` module existed solely to answer one question: are we running on Windows?

That single-method module was more ceremony than it was worth, and it meant platform detection was split across two places:

1. [`Win32.windows?` → in win32.rb](https://github.com/ruby/rake/blob/d9f85ffd9412df0175ec66ba28d682b40c8f3914/lib/rake/win32.rb#L11-L13) and
2. [`Application#unix?` → in application.rb](https://github.com/ruby/rake/blob/d9f85ffd9412df0175ec66ba28d682b40c8f3914/lib/rake/application.rb#L386-L389).

Both methods now live together as `Rake.windows?` and `Rake.unix?` on the Rake module itself, which is the natural home for gem-level helpers.

`Application#windows?` and `Application#unix?` are kept as thin delegators so nothing calling them through an application instance breaks.

Beyond tidying up the code, the refactoring also untangles two distinct uses of platform detection that were previously conflated. :tada:

The first is a static, load-time concern: deciding which tests to define or which assertions to run. This only needs a simple module-level predicate — call it once when the class loads and you're done.

The second is a dynamic, instance-level concern: branching on platform inside `Application` methods, and — critically — being stubbable in tests so that platform-specific code paths can be exercised regardless of where CI runs.

These two uses have different requirements.

1. The new `Rake.windows?` / `Rake.unix?` module methods serve the first cleanly.
3. The delegators on `Application` preserve the second — including the ability to stub `@app.windows?` / `@app.unix?` in tests, which several existing tests rely on.

I hope you like it, @hsbt 🤞 - this isn't just cleanup, it's a separation of concerns that makes the intent of each call site clearer.

---

[released](https://released.blabberate.com/p/ruby/rake/729)


test/test_rake.rb
```diff
@@ -38,4 +38,11 @@ def test_original_dir_reports_current_dir
     assert_equal @tempdir, Rake.original_dir
   end
 
+  def test_platform_detection
+    # ensure that Rake's platform detection logic
+    # is mutually exclusive, and doesn't claim to
+    # be both Windows and Unix at the same time
+    assert ! (Rake.windows? && Rake.unix?)
+  end
+
```
pvdb, line 47: This preserves the original `TestRakeApplication.test_windows` exclusivity check, but - arguably - with a more intent-revealing name, and also by documenting it _(as the original assertion was somewhat esoteric)_ 😃

PR #737, opened 2026-05-19
Add support for test-unit loader

tikkss:
Since test-unit 3.6.8 supports the `test-unit` command. And the `testrb` command is no longer available.


hsbt (approved):

tikkss:
Thanks for your review!

